PR #50, opened 2015-01-20
Tab sensitive load event handling.

tuxor1337:
It frequently occurs to me that the "Goto, Fill And Submit"-feature doesn't work as expected. It will only wait for _any_ of the currently open tabs to load and then fill in credentials for that particular tab. That's of course problematic whenever there are several tabs loading at the time when the "Goto, Fill And Submit" is fired. 

The proposed patch is far from the most elegant solution and it might still be problematic in some cases. But it works for me. (Especially, because I'm not using the preference "Try to auto fill page forms".)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -3,5 +3,6 @@ TARGET = ../bin/passff.xpi
 all: xpi
 
 xpi:
+	mkdir -p `dirname $(TARGET)`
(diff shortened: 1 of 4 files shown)


jvenant:
Looks great for me. Much much better approach than mine.
Thank you very much :wink:

PR #57, opened 2015-02-26
add info about problems on OS X

tophattom:


jvenant:
Thank you very much @tophattom

PR #62, opened 2015-04-16
Fix doubled menubar entry

innir:
For me the add-on only works with this change.


jvenant:
Oups ^_^
Didn't saw this.
Thank you very much

PR #67, opened 2015-06-07
Use placeholder for search instead of a label

sometoby:
A small UI improvement that removes the search label and replaces it with a placeholder in the search field.


jvenant:
Nice improvment.
Thank you very much for your contribution

PR #79, opened 2015-06-23
add --login as default parameter

phryneas:
when using "through shell" mode


jvenant:
Thanks. Hope it will avoid some issues creation  ;)

PR #80, opened 2015-06-23
Compact view

horosgrisa:
Compact view for refresh and options buttons


jvenant:
You're right : one line is enough for 2 buttons ^_^.
Thank you very much

PR #83, opened 2015-06-23
Add JS code conventions, reformat source code to comply, general code cleanup

sometoby:
Lots of changes, testing is encouraged.

The most important things to note are:

1. I added a few jshint annotations and a .jshintrc. If your editor supports jshint, using it is encouraged.
2. I removed the license header in ```bootstrap.js```. Since I found what I believed to be the original source file and saw that it had massively changed since then, I don't think the license header is necessary anymore.
3. Please, please use jshint. It will help avoid mistakes like the one fixed in 58c5a82.

I did a few tests with these changes applied and didn't notice anything massively broken. It would probably be a good idea to introduce a few test cases to prevent regressions, though.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -1,0 +1,64 @@
+# PassFF Javascript Conventions
+
+Please follow these conventions and amend to them when necessary.
+
+## Indentation
+
+Two spaces.
+
+## Line length
+
+At most 90 characters.
+
+## Semicolons
+
+Always terminate statements with semicolons.
+
+## In-line functions
+
+Please don't.
+
+```
+// good
+toggleKey.addEventListener('command', function(event) {
+  event.target.ownerDocument.getElementById(PassFF.Ids.button).click();
+}, true);
+
+// bad
+toggleKey.addEventListener('command', function(event) {event.target.ownerDocument.getElementById(PassFF.Ids.button).click();}, true);
+```
+
+## Braces
+
+On the same line as the statement.
+
+```
+// good
+if (true) {
+  // your code here
+}
+
+// bad
+if (true)
+{
+  // your code here
+}
+```
+
+## Line breaking
+
+When breaking statements into multiple lines:
+
+* Place function arguments one character to the right of the opening paranthesis.
+```
+PassFF.Menu.createContextualMenu(aBrowser.ownerDocument,
+                                 aBrowser.ownerGlobal.content.location.href);
+```
+
+* Break chained function calls before the connecting dot, and align the dot
+  with the last dot on the previous line.
+```
+let domWindow = aXULWindow.QueryInterface(Ci.nsIInterfaceRequestor)
+                          .getInterface(Ci.nsIDOMWindow);
+```
+
--- a/src/.jshintrc
+++ b/src/.jshintrc
@@ -2,9 +2,16 @@
   "moz": true,
   "globals": {
     "AddonManager": false,
+    "Cc": false,
+    "Ci": false,
     "Components": false,
+    "Cu": false,
     "CustomizableUI": false,
     "PassFF": true,
-    "Services": false
+    "Promise": false,
+    "Services": false,
+    "URL": false,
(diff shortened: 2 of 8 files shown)


jvenant:
I have 2 issues currently with those changes : 
* Global shortcut doesn't work any more
* Addon don't work anymore (no icon) after a firefox restart

sometoby:
Found the cause for the shortcuts breaking, will push a fix later.

As for the other problem, I'm afraid I can't reproduce it. Could you post some error logs or give a more detailed description of how to trigger the problem?

jvenant:
Ok, I gonna investigate a little bit deeper

jvenant:
Sorry, it works now. Don't know what happen this morning...?

jvenant:
Wait... don't work once again....
This bug is going to be funny I think ^_^
Not working logs : 
```
console.debug: 
  [PassFF]
  init
console.debug: 
  [PassFF]
  Wait documents
console.debug: 
  [PassFF]
  Wait documents
console.debug: 
  [PassFF]
  Documents found
```

Working logs

```
console.debug: 
  [PassFF]
  init
console.debug: 
  [PassFF]
  Wait documents
console.debug: 
  [PassFF]
  Documents found
console.debug: 
  [PassFF]
  Add panel to new window
console.debug: 
  [PassFF]
  createContextualMenu
  about:addons
console.debug: 
  [PassFF]
  Search items for:
```

I think it is linked to how fast your computer is (and how fast your Firefox start).
I remember I had to deal with this waiting for some documents to be created before launching ui creation.
It's inside waitForDocuments in bootstrap.js

jvenant:
I'm not 100% sure but in boostrap.js, I change all the
```
notify: PassFF.waitForXXXXX
```
to
```
notify: function() { PassFF.waitForXXXX() }
```

And for now, I didn't reproduce the bug

sometoby:
That's a bid odd, those statements should be equivalent.

sometoby:
Maybe being called in a different context breaks the waitFor methods. I'll revert that change, then.

jvenant:
From what i remember it is not : 
In javascript, your context (this) is the object who call the method.
so for 
```
notify: PassFF.waitForXXXXX 
```
your calling object is the timer
and for 
```
notify: function() { PassFF.waitForXXXX() }
```
your calling object is PassFF

Might be approximative, but it's something like that

jvenant:
I love the way javascript manage  stack heap and namespace..... :(

sometoby:
It's very likely the context problem. I hacked up an example and tried to approximate the situation.

https://jsfiddle.net/zjug8mgs/

A wrapper function solves this problem, but a bind call would too. Which one do you find easier to understand?

phryneas:
@jvenant it's actually quite logical - and I think you got it right, although I'd have to touch up on that to say it for certain.
I can recommend reading "Secrets of the JavaScript Ninja". But sorry, I'm off-topic ;)

phryneas:
Oh, and I would vote for the bind call. Looks a bit weird, but looking back at the code it will trigger the "yes, we did that for a reason and should not refactor it" effect.

sometoby:
Replaced the wrapper functions with a bind call. I think we're ready for a merge. @jvenant, how about it?

jvenant:
Didn't see anything else. It's OK for me too

jvenant:
Done

jvenant:
Oups, forgot the shortcut fix. @sometoby do you have it somewhere ?

sometoby:
The shortcut fix is already in master.

https://github.com/jvenant/passff/blob/master/src/modules/preferences.js#L26

jvenant:
Sorry, I'm not talking about this one. I'm talking about [this one](#issuecomment-115182108)

sometoby:
That was fixed in 6880303, is it still broken?

sometoby:
Apparently, removing the oncommand attribute breaks it. I reverted the change and pushed it to master.

PR #86, opened 2015-06-23
add german translation

phryneas:


sometoby:
Thanks! I tried my hand at a German translation, but you did a much better job than me.

phryneas:
By the way, is there some kind of special editor to see the languages side-by-side? Vim is okay for the job, but as soon as strings get added/dropped I think it will get quite messy.

sometoby:
I'm not aware of any editors or tools specifically for Firefox addon localization. We could of course hack up a Vim plugin or a shell script to check these things, but the best bet is probably just doing a quick ```grep -r``` to see if the locale key you're modifying is still being used somewhere.

phryneas:
Well, that would work, too xD

PR #87, opened 2015-06-24
Remove unused locale string. Fix russian translation

horosgrisa:


sometoby:
Thanks. You might want to consider not immediately sending a PR of your changes and reviewing them twice or so.

Issue #63, opened 2015-04-26
UTF-8 characters not displayed correctly


bricewge:
If some entries contain UTF-8 characters such as accented letters they are not displayed correctly.
For example `Université` become `Universit\303\251`.

jvenant:
Does it work when you use pass script directly ?

bricewge:
Yes, using pass from the command line correctly display UTF-8 characters.

jvenant:
I can't reproduce. I tried using accents and Russian characters, everything works.
I think it's something about your Firefox execution environment.

Could you try using the "Through shell" approach please (in the preferences)

bricewge:
I tired the "Through shell" option but it doesn't change anything. The problem is not only graphical because I can't acces the password under directories with accents, when ask to displayed it the password field is empty. The other directories, without fancy characters, aren't affected.

Here is a screen-shot of the problematic directory:
![passff](https://cloud.githubusercontent.com/assets/5525646/7386916/330ba106-ee59-11e4-9628-aff321b4589e.png)

I'm using firefox 37.0.2 under a up-to-date Archlinux.

jvenant:
hmm, that's weird. I'm pretty sure this is about your environment. Could you also try launching Firefox from your command line ? the same one you use to execute pass directly.

bricewge:
Launching Firefox from the command line – the same which I use to acces pass – doesn't change anything, nor using "Through shell" or "Direct call". Unfortunately I had to reinstall my box from scratch this weekend, so it doesn't look to me this issue come from some custom config I did since it's still present in my new install.

I'm usually using zsh, but it also work with bash, here is a dump of my [env](https://gist.github.com/bricewge/c71cc7973ba5b4214bb6).

jvenant:
Could you also activate the debug mode and give me output from the web console for PassFF please

Thank you

bricewge:
I just tried to debug the addon, but I don't know how to, sorry. The console output nothing, but it came from me as it's the same with my others addons.

Here is the return for the ls
```
drwx------ 1 1000 3013  78 mars  29 20:05 Université
```

jvenant:
Sorry for the delay, I didn't see your answer.
To enable debug mode, just go in the addon preferences and check the "Enable logs" check box at the bottom of the dialog box. You then have to open the Browser console (Tools/Web Developper/Browser console)
You should see many lines about passff. You could Filter on "[PassFF]" if you want

bricewge:
[...]

PR #88, opened 2015-06-24
Changes in interface

horosgrisa:


sometoby:
Could you please elaborate more on what this change is intended to do?

horosgrisa:
Small change: type of search box now is 'search', and removed menu separator.
2015-06-24-174903_3046x1050_scrot

sometoby:
Changing the type of the textbox to 'search' does not seem to actually do anything. Is there a good reason for the change? According to MDN, all this does is make the textbox emit the command event when keystrokes happen. Changing the type would only make sense if we were listening for the command event, which we are not.
As for removing the separator, @jvenant has the ultimate authority on visual changes.
I would also like to remind you once more to review your changes thoroughly before sending pull requests. As a maintainer, I get notifications about issue and PR updates on this project, and receiving multiple emails for one PR gets a little annoying.

sometoby:
I'd like to merge #83 sometime soon. It'd be great if you could put the separator back so I can merge this change and then rebase my own changes onto the new master. If the removal of the separator is really important to you, please send another pull request for @jvenant to review.

Also, for small changes like this, a simple issue would be sufficient. It's much easier for everyone involved when no one needs to do unnecessary rebases/merges. Once a solution has been agreed on, a maintainer can just make the change in the master branch.

jvenant:
Transforming the input type from text to search make sens I think because semantically, it is a search field. So even if we don't use the command event, I think it goes in the good way. (And I like the search icon ^_^)

For the separator, removing unnecessary elements makes the interface simpler so it's ok too

I'm a little bit more concerned about changing the "keyup" listener to a "input" listener because the behavior of those two events slightly differ. Especially concerning special keys like arrows, shift, ctrl...
Possibly, removing the "if(event.keyCode <= 46) return false;" lines may be enough. We have to check that.

But any way, making @sometoby refactoring/clean up work easier is currently our priority. So testing PR #83 to merge it asap on the master must be done first. We'll see for this PR after.

Thank you for your contribution @horosgrisa. I really appreciate. We will integrate it just after.

horosgrisa:
@jvenant, I found a bug: I have Cyrillic files and passff doesn't shows names correctly
2015-06-24-234343_1366x768_scrot

sometoby:
@horosgrisa The bug is unrelated to this pull request. Please open an issue for this problem.

jvenant:
try using the through shell invocation with --login as shell argument. See #63 for details.
If it still doesn't work, please create an issue for this. PR is not the place for bug report

PR #112, opened 2015-10-30
fix formatting of password info

carlwgeorge:


sometoby:
Seems like people actually do read the manual. :wink: 

Thank you for the fix!

PR #119, opened 2016-01-06
Do not add items to the menu if they are not leafs

craftyguy:
This resolves https://github.com/jvenant/passff/issues/115, where auto fill was incorrectly using the directory name as username


jvenant:
Awesome. Make sense.
Thank you very much

PR #132, opened 2016-02-07
Clarify README

oneirosFade:
After mucking about reading old issues and trying to find out why passff wasn't behaving as expected, I found that the "Issues with OS X" section is actually relevant to non-OSX platforms, so I changed the README to reflect this. Also added a bit about installing from git, and fixed a couple minor things.


jvenant:
Thank you very much.

Issue #147, opened 2016-04-23
No preferences window in German translation


fdw:
Whenever I try to open the preferences window, I get an error:

    Not defined entity
    Address: chrome://passff/content/preferencesWindow.xul
    Line 77, column 17
    <label tooltiptext="&enter-behavior-tooltip;" value="&enter-behavior;"/>

I think the problem is with the German translation, as the entity enter-behavior-tooltip is not defined there.

fdw:
I created a pull request to fix this: #148

fdw:
Merged :)

PR #148, opened 2016-04-23
Fix for #147: Add strings to German translation

fdw:
This commit adds two strings to the German translation,
* enter-behavior
* enter-behavior-tooltip.
This probably fixes a bug where the preferences window cannot be shown, as the entities are not defined there.

Unfortunately I could not test it, because the addon needs to be verified before my Firefox will run it.


fdw:
Hi,
do you need anything else for the pull request?

jvenant:
No, just time ^_^

fdw:
Cool, thank you. Any idea when you will release a new version at Mozilla?

jvenant:
I've just submitted the new version. Because there is no big changes, I think it should be fast to validate.
Probably tomorrow...

jvenant:
You can try here for an early access : 
https://addons.mozilla.org/en-US/firefox/addon/passff/versions/
(version 0.1.31)

fdw:
Thank you so much!

jvenant:
Sorry for the delay ¯\_(ツ)_/¯

PR #164, opened 2016-09-11
Autofill for other inputs

bendikrb:
Allows for filling out inputs found in passwordData, which isn't login, password or url.


jvenant:
Awesome. Very useful feature.
Thank you very much

PR #167, opened 2016-10-13
Better defaults for OS X

### LINKED ISSUE #72: Sensible defaults for OS X
It would be more convenient if the pass script configuration defaulted to the known workaround on OS X.

nwallace:
Detects OS to provide "through shell" defaults for OS X (fixes #72).  Also works around a bug in pass (bug has been fixed upstream but still not released) where .gpg file extension is not stripped off correctly, which breaks this plugin.

### DIFF
--- a/src/modules/preferences.js
+++ b/src/modules/preferences.js
@@ -3,242 +3,262 @@
 
 const {TextDecoder, OS} = Cu.import('resource://gre/modules/osfile.jsm', {});
 
-PassFF.Preferences = {
-  _environment: Cc['@mozilla.org/process/environment;1']
-                .getService(Components.interfaces.nsIEnvironment),
-  _gpgAgentEnv: null,
-  _params: {
-    passwordInputNames    : 'passwd,password,pass',
-    loginInputNames       : 'login,user,mail,email,username,opt_login,log,usr_name',
-    loginFieldNames       : 'login,user,username,id',
-    passwordFieldNames    : 'passwd,password,pass',
-    urlFieldNames         : 'url,http',
-    command               : '/usr/bin/pass',
-    commandArgs           : '',
-    shell                 : '/bin/bash',
-    shellArgs             : '',
-    home                  : '',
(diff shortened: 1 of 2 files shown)


src/modules/preferences.js
```diff
@@ -3,242 +3,262 @@
...
-  },
+    get passwordInputNames() {
+      log.error("aaaaa ")
+      return this._params.passwordInputNames.split(',');
+    },
 
-  get passwordFieldNames() {
-    return this._params.passwordFieldNames.split(',');
-  },
+    get loginInputNames() {
```
nwallace: You know I just noticed that the pass.js file [has some `let`s at the top].  Are those scoped to just that file?  I don't know enough about how the variable scopes work in extensions, but I could skip the closure that created the whole whitespace issue if the `let` to define `getDefaultParams` doesn't leak outside this file.  What do you think?

phryneas: Personally I think that those files are scoped like modules, but I'm not really sure about that, as I have just been writing smaller patches myself. 
Also, I don't think that `let` should ever leak into the global scope - Mozilla states "At the top level of programs and functions, let, unlike var, does not create a property on the global object."
@jvenant, do you have more input on that?

nwallace: @jvenant Also, to your initial question about the ordering -- I believe the order of everything is the same -- it looks like GitHub's diff here just didn't do a good job of recognizing that.  If you look at the files by themselves, you can see order is the same in the original version and in this pull request:  ..., passwordInputNames, loginInputNames, loginFieldNames, passwordFieldNames, urlFieldNames, etc...

phryneas: argh, I see it - github diffs without taking whitespace into account and moved everything down by two properties. sorry -.-

jvenant: Hi,

Thank you very much for your patch.
Unfortunately I'm working today. I'll review and merge it asap.

PR #168, opened 2016-10-20
Add passwords

nwallace:
This lets you add passwords right from the extension. You can specify the password yourself, or generate one. After the new password is added, the password list will be refreshed so that the user can find it in the extension.  Some new preferences let you specify your preferred default settings for adding new passwords.


nwallace:
I should point out that I only included English translations for the new strings used.  I don't know Russian or German, but I could use Google Translate to fill them out

PR #174, opened 2016-11-19
Show URL on Display

gutierri:


nwallace:
Thanks for the contribution! This is a great idea. However, right now if you don't have a URL in your password file, it displays "URL: (null)". I'd rather it leave off the whole URL line if there's no URL in the file, or at least say "URL: " without the "(null)".  Could you make that work?

gutierri:
@nwallace OK!

gutierri:
@nwallace Ready!(?)

PR #180, opened 2016-12-27
Search through frames for credential fields as well, not just iframes

GordanV:
This pull request adds support for login forms in \<frames\>.


tuxor1337:
@GordanV Thanks for your suggestion! It was very easy to adapt it to the current WebExtensions based code :)

PR #181, opened 2017-01-11
Hotfix/undef entities

k127:
There seems to be work in progress on [nwallace/passff@master](https://github.com/nwallace/passff/tree/master)

This is just a hot fix to keep Firefox from complaining.


tuxor1337:
Since the switch to WebExtensions on the master branch, this is not relevant any longer.

Issue #169, opened 2016-10-22
Migrate to WebExtensions


emersion:
See https://developer.mozilla.org/en-US/Add-ons/WebExtensions

jvenant:
ouch ! Small ticket, but it involve to redesign almost everything.
I already check a little bit webextensions.
My main concern is about executing external program. It seems to require a much more complex install process.

emersion:
Yes, this involves rewriting a lot of code, but the WebExtension version would be compatible with all major browsers (Firefox, Chrome, Opera, Edge).

I see two ways of doing it:
* Using native messaging: https://developer.chrome.com/extensions/nativeMessaging
* Using the `fileSystem` API: https://developer.chrome.com/apps/fileSystem

emersion:
Damn, the `fileSystem` API is only for Chrome apps, it's not part of WebExtensions.

hallabro:
I would be willing to work with somebody to migrate this extenstion to WebExtensions if this is the prefered long term solution. I'm no expert my any means but I know my way around in javascript.

Currently I am in the position of the extension not working because I use e10s without the shim and would very much like to have a working extension. Manually copying from pass is no fun. ☹️

emersion:
There's one other solution : use a pure JavaScript Git implementation (e.g. https://github.com/maryrosecook/gitlet) + OpenPGP.js.

hallabro:
Well, `pass` may be used without git so I guess it is no universal solution.

What would be the reason to use git? To get a list of files to be decrypted by OpenPGP.js?

emersion:
A WebExtension has no way to interact with the filesystem - so we can't directly read the local pass database. Using the nativeMessaging API would require to install pass and another binary.

hallabro:
One other idea on top of my head would be to have a pass daemon providing a local WebSocket which could talk to (which I think is how 1Password does).

emersion:
Either way, something else needs to ne installed. I think it's better to use nativeMessaging since there wouldn't be a deamon always running in background.

hallabro:
Yes, I think you are right in using Native Messaging, Could you please elaborate in what other binary would be required for Native Messaging? I understand that the pass binary would be required.

emersion:
To be able to connect to an external program from a WebExtension, we need:
1. To put a manifest file (for Firefox on Linux, it's in `~/.Mozilla/native-messaging-hosts/<name>.json`
[...]

PR #196, opened 2017-05-04
Make install script easier to run

nwallace:
This commit changes the host app install script to download the host app and manifest files from GitHub instead of using files on disk.

As discussed in #169 , I feel like the install script should be very easy to run. This change enables a user to do one-liners like this to install the host app:

```
$ curl -sSL https://github.com/nwallace/passff/releases/download/1.0/install_host_app.sh | bash -s -- firefox
```

Note: Currently the script downloads the files from the master branch on GitHub. Before a release, we should update this to refer to a tag or release file in GitHub instead.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -30,18 +30,25 @@ Clone the repository. Then, from the project's `src/` directory, execute `make`.
 For the extension to communicate with your system's `pass` script, you need to install what's called the host application.
 
 #### Official release
-Download the `install_host_app.sh` script from [our releases page](https://github.com/nwallace/passff/releases) and execute it.
+Download the `install_host_app.sh` script from [our releases page](https://github.com/nwallace/passff/releases) and execute it. You can do this in one line like so:
+
+```
+$ curl -sSL https://github.com/nwallace/passff/<VERSION>/install_host_app.sh | bash -s -- [firefox|chrome|opera|chromium|vivaldi]
+```
+
+This script will download the host application (a small python script) and the add-on's manifest file (a JSON config file) and put them in the right place.
+If you're concerned about executing a script that downloads files from the web, you can download the files yourself and run the script with the `--local` option instead or link the files yourself. Details below.
 
 #### Latest from GitHub
-Clone the repository. Then, from the project's `host/` directory, execute the installation script for your desired browser (`--firefox`, `--chrome`, `--opera`). Omit the browser flag to install it for any browser you have on your system.
+Clone the repository. Then, from the project's `host/` directory, execute the installation script for your desired browser (`firefox`, `chrome`, `opera`, `chromium`, or `vivaldi`).
 
-  ```
-  $ ./install_host_app.sh --firefox
-  ```
+```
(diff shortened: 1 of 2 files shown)


nwallace:
I [just added an option](https://github.com/nwallace/passff/pull/196/commits/331da1c5348e369b0aff19feec24b602210b83b9) to the install script to let the user specify to use local files instead of downloading them from the internet. I hope this will abate any fears users might have about a script downloading files on their behalf. It will also be handy for development purposes in case a contributor needs to modify the host app or manifest file.

Any thoughts on this? @yurikhan -- you raised this concern. Does this help?

yurikhan:
The option will be handy, yes.

As for the download URLs: You are correct that referring to a branch is unsafe. As the branch moves with each commit, results of a review would be invalidated almost instantly. Tags can be easily moved, too. [Releases can be edited](https://help.github.com/articles/editing-and-deleting-releases/). An attacker who compromises a developer account will be able to tamper with the installation script and/or messaging host and execute arbitrary code on the user’s machine with user privileges.

Referring to a specific commit hash is safer in this regard, as long as GitHub itself is not compromised and SHA-1 is not broken.

phryneas:
hm, this could be verified using signed commits/tags? using a sparse checkout, it should be possible to check out only what is necessary
of course, this requires git to be installed and introduces another dependency

```
mkdir install-host
cd install-host
git init
git config core.sparseCheckout true
echo src/host/passff.py > .git/info/sparse-checkout
git remote add -f origin https://github.com/nwallace/passff
git fetch
git tag -v 1.0.0-alpha || exit 1
git checkout tags/1.0.0-alpha
```

phryneas:
Also, this introduces an additional requirement for git and the signing key should somehow be trusted.
This could be offered as an alternative form of installing for the paranoid that are okay with the additional hassle of importing the key.

tuxor1337:
As far as I understand it, `git` is not an *additional* requirement insofar as `pass` is already required and `pass` requires `git`.

tuxor1337:
@yurikhan You are concerned about compromised dev accounts, right?
* Since the addon will usually be delivered through addons.mozilla.org and you can't upload additional files there, there will have to be host app installation instructions on addons.mozilla.org pointing the user to a third party source.
* Let's assume it's unlikely that the developer account on addons.mozilla.org will be compromised because it's only a single person. Opposed to that, it's likely that one of the dev accounts having access to the Github repo will be compromised.
* Why not include some kind of hash into the installation instructions on addons.mozilla.org then?

nwallace:
@tuxor1337 I agree. I think providing some kind of hash is about the best we can do for the official browser extension releases you get from addons.mozilla.org and such.

As for the host app installation, there are instructions for installing it yourself manually here: https://github.com/nwallace/passff/blob/331da1c5348e369b0aff19feec24b602210b83b9/docs/INSTALLATION.md#latest-from-github-1

The security-conscious can install them manually and read the files to make sure there's no funny business going on (they're very small, so anyone could understand them).

I'm satisfied with this approach, and I think we've got some good alternatives if anyone has concerns with it.

Issue #202, opened 2017-06-04
Windows: Host App installation process not working.


phryneas:
As the title says: At least Chrome in windows needs a registry entry, also the shebang alone in the .py file might not be enough -  maybe something like a .bat wrapper could work.

I can do this within the next month or so if you need someone to implement this.

nwallace:
That would be great. I don't have access to a Windows machine, so I can't do this myself.

I assumed everyone would know the Windows installation process wouldn't really work... how does pass itself work on Windows? Don't you need to be running Cygwin or the new Linux Subsystem anyway?

phryneas:
Pass itself works great with most types of bash compiled for windows, which would include the Linux Subsystem for Windows, Cygwin, Msys2 or the Gnu tools for windows.

I'm a linux-windows mixed user, so I always have a nicely working console on my windows machines ;)

As for passff - the old passff worked using the "though shell" method, bit it had it's quirks lately with the resetting configuration. So updating to HostApp would not be the worst idea for me ^^

I'm currently writing a HostApp & Typescript library for Chrome/Firefox to use the native GPG from the browser without having to store private keys within the browser, so I think this will work quite well, although I'll have to look into the python part. I had been using go, so there was no need to specify an interpreter.

gutierri:
close (ref #203 ,  cc @tuxor1337 )?

phryneas:
works fine on two of my windows machines, so I gues you can close it

Issue #201, opened 2017-06-02
Override $GNUPGHOME in preferences


mem-:
As a complement to issue #96 I'd like to have the possibility to override my default $GNUPGHOME settings.

Background:
I use $GNUPGHOME pointing to a specific location for my regular keyring.
For pass I created a new (dedicated) keyring, stored in an other path.
For regular pass usage, I've set  PASSWORD_STORE_GPG_OPTS="--homedir some-alt-path"

Possible solutions:
- Use $GNUPGHOME as today and apply $PASSWORD_STORE_GPG_OPTS when applicable
- and/or add a preferences setting in PassFF for its "internal" $GNUPGHOME value

Regards,
// Mem

tuxor1337:
You can now set arbitrary environment variables in the preferences. This should fix this issue.

PR #203, opened 2017-06-05
create batch file for installation in windows

phryneas:
See #202.

This pull request adds a windows batch file to install the Host Application with the same (but only long -- parameters) as the bash script.

Files are installed to `C:\Users\USERNAME\AppData\Roaming\passff`. Next to the passff.py and passff.json which will either be downloaded or installed from the current directory, a `passff.bat` will be created with the following contents:

```
@ECHO OFF
SET "PATH=<value of PATH at the time of installation>"
SET "GNUPGHOME=<value of GNUPGHOMEat the time of installation>"
python "C:\Users\USERNAME\AppData\Roaming\passff\passff.py"  %*
```

Setting PATH this way is a necessity for python to be found in most windows installations, and if the installer file is called from a shell that has access to the `pass` command, that is already configured correctly as well.

I added the GNUPGHOME variable to satisfy #201 at least in Windows. 

The approach of a wrapper-script that inherits variable values from time of installation might generally solve many issues with wrongly set PATH and other environment variables. 
This might be something we could consider for the linux side as well - we might even be able to remove big chunks of code from the settings & host application that way.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -39,6 +39,16 @@ $ curl -sSL https://github.com/passff/passff/releases/download/1.0.0/install_hos
 This script will download the host application (a small python script) and the add-on's manifest file (a JSON config file) and put them in the right place.
 If you're concerned about executing a script that downloads files from the web, you can download the files yourself and run the script with the `--local` option instead or link the files yourself. Details below.
 
+#### Windows
+Download the `install_host_app.bat` script from [our releases page](https://github.com/passff/passff/releases) and execute it from within a shell with a correct PATH.
+*The rule of thumb is: if you can execute pass and python from your shell, then your host application will be installed correctly.*
+
+```
+> install_host_app.bat [firefox|chrome|opera|chromium|vivaldi]
+```
+
+Note: Older Windows versions might require powershell to be installed manually as the install script uses powershell internally. Windows 10 users should be fine out of the box.
+
 #### Latest from GitHub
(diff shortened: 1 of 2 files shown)


nwallace:
Awesome, thanks! Could you also update the installation instructions in `docs/INSTALLATION.md`?

Also, since I don't have access to a Windows box, can anyone else test this to confirm it works on others' systems?

phryneas:
Sorry, I was on vacation for a few days. Here are the updated instructions.

docs/INSTALLATION.md
```diff
@@ -39,6 +39,16 @@ $ curl -sSL https://github.com/passff/passff/releases/download/1.0.0/install_hos
 This script will download the host application (a small python script) and the add-on's manifest file (a JSON config file) and put them in the right place.
 If you're concerned about executing a script that downloads files from the web, you can download the files yourself and run the script with the `--local` option instead or link the files yourself. Details below.
 
+#### Windows
+Download the `install_host_app.bat` script from [our releases page](https://github.com/passff/passff/releases) and execute it from within a shell with a correct PATH.
+*The rule of thumb is: if you can execute pass and python from your shell, then your host application will be installed correctly.*
+
+```
+> install_host_app.sh [firefox|chrome|opera|chromium|vivaldi]
```
gutierri: The correct would not it be so? _install_host_app.**bat**_

phryneas: Damn, you're abosolutely right. I'll amend this.

gutierri:
cc @nwallace It's request ok for merge

tuxor1337:
I don't have access to a Windows machine for testing, but, as @gutierri said, I guess it's okay for merge.

PR #204, opened 2017-06-10
port to typescript

phryneas:
To preface this: I am not sour in any way if this is not accepted. I just had fun porting passff to typescript.

Now, that that's out of the way:
I ported the current passff master to typescript.
Almost every variable and method parameter has a definitive type now.

In the process of doing this conversion, I found some places where values where passed that should never be passed, or where values were checked for types that never were passed. 
Meaning, I found a few bugs. These are pointed out as //TODO comments and I fixed them to the best of my knowledge.
Even if this is not going to be accepted, those bugs should be fixed.

I did my best to keep line association in the diffs. Of course, that was not possible everywhere, but I hope it's readable. (pass.js -> pass.ts can be compared in my phpStorm, but github does not want to recognize it as the same file)

To run this, you have to do `npm install` followed by `npm run build`. If you want to watch the files for changes and automatically build a new .xpi, you can run `npm run watch`.

If this is actually going to be accepted, I'd happily write a configuration for something like Travis CI, so we have automatic build, and pull requests can automatically be checked for type correctness.


src/modules/pass.ts
```diff
@@ -0,0 +1,588 @@
...
+  static getPasswordData(item: Item): Promise<Dict<string>> {
+    let result: Dict<string> = {};
+
+    if (item.isLeaf()) { // multiline-style item
+      let args = [item.fullKey()];
+      return this.executePass(args).then((executionResult) => new Promise((resolve, reject) => {
+        let gpgDecryptFailed = executionResult.stderr
+            .indexOf('gpg: decryption failed: No secret key') >= 0;
+
+        while (executionResult.exitCode !== 0 && gpgDecryptFailed) {
```
phryneas: Oh, I also stumbled over this. 
I guess that could result in an infinite loop and should be removed, shouldn't it?

gutierri:
Is that really necessary? I see it's another step to complexity in this case.

phryneas:
@gutierri: in my eyes, this could be a first step towards a more readable code base.

Right now, passff is partially written in a very mozilla-specific Javascript-Dialect (there is use of things like `for (var [name, value] in Iterator(a))` which at one point was valid, but is called deprecated even by Mozilla as it is superseeded by ES6's `for ... of` loop ).
The whole plugin is essentially one big Object `PassFF` with sub-objects scattered among different files.
Scattered next to it is one "Class" (Item) using the Javascript Prototype feature.
As objects need to be passed between the GUI context and background context, they are converted to more simple object with sometimes slightly different property names.
Somewhere, even `String.prototype` gets extended, with not even a real benefit (calling a normal function at 5 or 6 different positions in the code has the same effect).

All that is not really bad in itself, but it alleviates developer error at every step.

What I've done in this pull request is only a first step of making this more developer-friendly:
* The Objects `PassFF`, `PassFF.Menu`, `PassFF.Page`, `PassFF.Menu`, `PassFF.Pass` and `PassFF.Preferences` are now separate classes that can be imported separately where required. 
* all properties of said classes now have fixed types, and if they were more complex structures, those types now have an interface that define each of their properties.
* all methods have fixed parameter types, somethimes with default values for optional parameters and only one type of return value.

This typization enables a much better experience in most modern IDEs (PhpStorm, Visual Studio Code, I also believe Netbeans) and you will get warnings if you accidentially pass a wrong type anywhere.

From that step on, refactoring parts of the code gets much easier, as one has clear knowledge, which code is called where and with which parameters.


That said: adding classes could have been done without typescript - Firefox natively supports a lot of the ES6 featureset by now. All that types could have been designed in JsDoc.
That definitely is a possibility too.

But it's about the same amount of work to go the full way instead of only half the way. And it would be relying of IDE+JsDoc typechecking, which is even in the best case far worse than the Typescript compiler.

nwallace:
Wow, thanks for the effort, @phryneas! I definitely agree with you motivations for this work: the current code is a bit of a mess. Lots of the code dates back several years, it has been written and rewritten several times, and it comes from people with lots of different coding styles. This makes for a codebase that is hard to understand and difficult to extend with new features. 

Type safety is definitely one way to help some of these problems, but I don't think it addresses the fundamental issue of our messy codebase. Personally, I tend to prefer unit tests and integration tests over type safety (and I'm planning on introducing some unit tests very soon). Finally, I worry that switching to TypeScript (or any other transpiled language) would limit contributions from the community by raising the barrier to entry.

So, unfortunately for your work @phryneas, I don't think I'm going to merge this in.

That being said, I'm definitely going to look through your comments where you identified bugs and make sure those get fixed or eliminated. I also think we should be using modules like you do here.

I've got my own refactoring going on [here](https://github.com/passff/passff/tree/beautification/rewrite) that will probably eliminate the bugs you found here, but your work here will help me make sure it does. Also, I should start using proper modules for dependencies.

Does that sound fair? Any objections?




For these reasons, I started [my own refactoring](https://github.com/passff/passff/tree/beautification/rewrite) a couple of weeks ago. The new version makes more consistent use of ES6, gets rid of several bugs, and is much better organized.

phryneas:
Hey Nathan,
I already expected that response - especially after I saw your beautification-branch (the day after I finished this - d'oh ^^). But I thought, I'd give it a try ;)

One thing about transpiled languages, or not using them, though: 

The hurdle of using them is extremely low, as it's all abstraced away in the package.json and grunt/browserify/webpack config that is already present. (but I admit, writing those is a pain in the ass)
It comes down to `npm install` and `npm run build` for the complete process of packaging an xpi.

As most other Javscript projects have some dependencies that have to be installed using `npm install`, this process will be something very familiar for most Javascript developers.

tuxor1337:
Thanks a lot, @phryneas - your comments and improvements will definitely be of invaluable use. When I converted this extension from the XUL-based version, I reused as much of the JavaScript code as possible - firstly, in order to keep my own workload small and, secondly, to make sure I'm not accidentally changing or removing any of the existing features.

But even before the change to WebExtensions, there existed a branch that wanted to rewrite large parts of the code base - and that was for a reason. When I was working on the switch to WebExtensions, it was obvious to me, that a major code review was advisable. And, in fact, nwallace started working on a refactoring very soon.

With this pull request, we now have at least two very thorough code reviews. But I agree with nwallace that we should rather implement robust unit tests instead of adding another layer of complexity.

phryneas:
Okay, that's one thing I can't refrain myself from adding here - mainly because I can absolutely see where you're coming from: Having seen this codes a year ago, I would have reacted the same.

_There's no real additional complexity there. That code is ES6, with some very minor sugar on top._

In the end, there's no real difference between that sugar and JsDoc.

Let me give you some examples:

```
class X {
public y : string = ""
}
```

comes down to ES6+JsDoc

```
/**
* @class X
* @property {string} 
*/
class X {
  constructor(){
    this.y = "";
  }
}
```
---

```
let x : string = "";
```

comes down to ES6+JsDoc

```
/** @var {string] x */
let x = "";
```
---
```
function x(a : string =  "", b : number= "") : string {
....
}
```

comes down to ES6+JsDoc
```
/**
* @param {string} a
* @param {number} b
* @returns {string}
*/
function x(a =  "", b = "") {
....
}
```
---
and 
```
interface MyType {
x: string
}
```
comes down to ES6+JsDoc
```
/**
* @typedef {MyType}
* @property {string} x
*/
```

There's not much more happening here than some (really necessary) comments in the code that tell developers that don't know the codebase very much what to expect.



Saying all that: I absolutely agree with all of you, that this can not be the only measure of increasing code quality and keeping the code quality up. 
Tests are a great idea.

Oh, and @nwallace: Can I reach you somewhere out of the issues? I think I can give you some feedback for the beautification branch, but I don't want to spam everyone that is watching the whole repository ;)

nwallace:
@phryneas Thanks! Yes, definitely feel free to email me using the email address on my GitHub profile

phryneas:
@nwallace - are you sure that's working? I have already sent you a mail regarding this pull request a few days ago, so I assumed that your mail address is dead ;)

PR #209, opened 2017-06-29
Fix computation of item quality

qsantos:
The former version tried to split the target domain name using a string
`'.+'` which cannot be part of a domain name. It caused the generation
of a single group and the quality became trivial. The new version
matches domain names as intended.

The former code had a trivial result:

    >>> hello.world".split('\.+')
    Array [ "hello.world" ]

In contrast:

    >>> hello.world".split(/\.+/)
    Array [ "hello", "world" ]


tuxor1337:
Looks fine. Thank you very much @qsantos !

Issue #105, opened 2015-09-25
Chrome Port


Immortalin:
Hi, can you make a port of this to chrome?

jvenant:
No, I can't.
I don't have even enough time to deal with all the issues of the Firefox version.
But if you want to implement a chrome/chromium version of the plugin, or if you know someone who want, you're very welcome ^_^

sometoby:
This may be relatively easy once the addon has been ported to use the addon API's. I suggest we leave this open for later evaluation.

SkyzohKey:
For people that are looking for a temporary replacement of PassFF over chrome, until the port is effective: [Browserpass](https://github.com/dannyvankooten/browserpass)

tuxor1337:
Now that we are on WebExtensions, it should be compatible with Chrome. Has anybody tried this yet?

nwallace:
I just tried it with Chrome today and it's complaining that our translation keys are invalid. Apparently Chrome doesn't allow dots in the names, only letters, numbers and underscores. Should be a quick fix, but I don't have time right now to test it. I'll get back to this

singron:
Chromium loads the extension with #213. I think it might only need the [webextension-polyfill](https://github.com/mozilla/webextension-polyfill) to run correctly (it fixes `browser` vs `chrome` namespace issues and Promise vs callback discrepancies).

5bentz:
I have randomly found another extension for chrome/chromium: [chrome-pass](https://github.com/hsanson/chrome-pass)

tuxor1337:
I had a look into chrome support using webextension-polyfill version 0.1.2 which looks good already. However, I did not get the native messaging host app running and ran into errors about "disconnected ports" - probably related to the browser messaging API.

Since I don't use Chrome or Chromium, I'm not willing to put more effort into this. However, if somebody else comes up with a working port, I will be happy to accept a pull request.

5bentz:
I randomly found that PassFF has this line https://github.com/passff/passff/blob/master/src/skin/menu.css#L119
which is only compatible with Firefox, and barely "supported" by other browsers. See [here](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance)

I tested PassFF's toolbar menu without it, but I failed to see any difference. What's the purpose of `-moz-appearance: none`?

Thanks!

tuxor1337:
[...]

PR #213, opened 2017-07-15
Rename translation keys to be compatible with Chromium

singron:
I just replaced `-` and `.` with `_`, which is the only symbol allowed.

This isn't sufficient to work in Chromium (#105). It still complains that 'browser is not defined'.


tuxor1337:
Thank you very much! As soon as I have tested it thoroughly, I'm willing to merge this 👍

PR #214, opened 2017-07-20
remove input type checking from other fields

mir06:
Hi. Unfortunately I have to use some web-pages that have two input fields of type password. 

passff will fill the password from the first line of the password file into **all** html input fields that have the password type - regardless of the presence of correct <other_inputfield_name>: <inputfield_value> lines. This is because isOtherInputCheck test for the loginInputTypes.

I removed this check and now the forms are filled correctly. I cannot see any side-effects. In fact, I do not understand why the other fields should only be text, email or tel inputs.

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -229,8 +229,7 @@ function isLoginInput(input) {
 
 function isOtherInputCheck(other) {
   return function(input) {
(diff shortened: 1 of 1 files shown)


tuxor1337:
Looks reasonable. Thanks for your contribution!

mir06:
Thx. and first of all thanks for that tool! 👍

PR #232, opened 2017-09-19
Improve support for forms that are not pure HTML and enhance submit button discovery

GordanV:
Certain websites disable submit/login buttons until all necessary fields
have been filled in. This commit fires an input event on each field
whose value has been set to notify the website about the change.

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -165,7 +165,7 @@ var passwordInputNames = [];
 var subpageSearchDepth = 0;
 
 function getSubmitButton(form) {
-  let buttons = form.querySelectorAll('button[type=submit]');
+  let buttons = form.querySelectorAll('button:not([type=reset])');
 
   if (buttons.length === 0) {
     buttons = Array.prototype.slice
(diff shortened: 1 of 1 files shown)


src/modules/page.js
```diff
@@ -176,13 +176,29 @@ function getSubmitButton(form) {
...
+    (button) => button.getAttribute("type") === "submit",
+    // the browser interprets an unset or invalid type as submit
+    (button) => !Array.prototype.includes
+                                .call(["submit", "button"], button.getAttribute("type")),
+    // assume that last button in form performs submission via javascript
+    (button, index, arr) => index + 1 === arr.length
+  ];
+
+  for (let predicate of submitButtonPredicates) {
+    button = Array.prototype.find.call(buttons, predicate);
```
tuxor1337: If I'm not mistaken, a variable declaration (`let button = ...`) is missing here.

GordanV: Oh, good catch, fixed.

tuxor1337:
Thanks! 👍

PR #241, opened 2017-10-27
UI improvements

RafaelCz:
Hello,
I hope it doesn't look too nitpicky, but once spotted, I can't unsee these things 😄 

The currently used icon is not optimized for screens with lower resolutions, which makes it look blurry on said screens. I recreated it by following [Mozillas design guidelines](http://design.firefox.com/photon/visuals/iconography.html), using a 16x16 pixel base and aligning all paths to the pixel grid, to provide a sharp looking icon on all sizes and resolutions.
Here is a comparison between the current and new icon (normal sized and zoomed view):
bildschirmfoto

As shown in the webextension example [beastify](https://github.com/mdn/webextensions-examples/tree/master/beastify), I have added a light version of the browser_action icon, that is going to be automatically used in dark themes:
preview1
Unfortunately this feature doesn't work flawlessly until Firefox 57+, FF 56 is using the wrong (light) icon in it's default theme. 

Furthermore I have replaced the icons used in the popup menu. Right now, the current preferences icon looks out of place, compared the the other two icons:
bildschirmfoto von 2017-10-28 00-12-12

By using icons [provided by the Mozilla UX-Team](http://design.firefox.com/icons/viewer/) the popup menu gets a consistent look:
preview2
These icons are licensed under MPL 2, but AFAIK it shouldn't cause a problem...


tuxor1337:
Thanks a lot! This was overdue, but I didn't have the time to find decent icons.

Issue #243, opened 2017-11-08
Firefox hangs when GPG fails to produce a password


shtrom:
When GPG or the agent is in such a state that it will output the following when trying to decrypt,

    gpg: public key decryption failed: General error
    gpg: decryption failed: No secret key

PassFF makes Firefox hang forever, and the only solution is to kill it, at which point it outputs the following to STDERR before dying.

    [GFX1-]: Receive IPC close with reason=AbnormalShutdown
    [Child 11678, Chrome_ChildThread] WARNING: pipe error (3): Connection reset by peer: file /builds/worker/workspace/build/src/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 353
    Killed

Save for the interesting reference to Chromium, it seems like it's hanging on an IPC, which I suspect would be with the PassFF helper, which itself is waiting for a decrypted stream that is not coming.

A timeout of some sort, or inspection of the return status from GPG in the helper might, well, help.

Is there a quick way I could run the helper manually (i.e., not through Firefox) to request a password decryption, so I can investigate further? Essentially, what command should I write into it?

tuxor1337:
By helper you mean the host application, right? Here, somebody modified it to log some debugging info to a local file: https://github.com/passff/passff/issues/219#issuecomment-340259566 Maybe this helps you at tracing down your issue?

drmoose:
This seems to be caused by [this infinite loop](https://github.com/passff/passff/blob/ae039624f6b136c8e6da643073952179e06042a3/src/modules/pass.js#L138).

tuxor1337:
Closed by #248

Issue #101, opened 2015-08-23
Quell "Can't access to your password" dialog when gpg-card is not inserted


alclary:
This is less of an actual _issue_ with the program and more of me seeking help on a customization. What are my options to disable the "Can't access to your password, Please add your passphrase to gpg-agent first" pop-up dialog which is initiated on pages on which passff auto-fills information. This message is only initiated when my gpg-card (containing my gpg-keys) is not plugged in. Is there a way to completely quell this dialog? Perhaps a conditional bash script would be possible if I knew how passff called upon and/or passed information to gpg-agent? Any idea where to start?

Thank you for any assistance or suggestions.

jvenant:
Currently there is no way to disable the dialog box. I will add an option to do that because I'm agree with you : most of the time it is annoying.
The dialog box appear if the sentence "gpg: decryption failed: No secret key" is found in the gpg/pass error output. So waiting for a new version including the new option, the only workaround I can see is to wrap the pass call in a script that would replace the sentence in something else or that would redirect the error output to dev/null.

tuxor1337:
Closed by #248

PR #248, opened 2017-11-19
Safer autofill

### LINKED ISSUE #220: Autofill: Asks for GPG password several times
Hi folks!

Started to use `passff` today, but TBH I'm not very happy about it yet. The installation of `pass` went fine and I use a private GitLab repo (via SSH), works.

Ok, now when I start FF, it asks me for the GPG password - I get that.

It asks me again about 5 seconds after that first one. And a third time about 30 seconds later.
It works, but I don't want to type my password 3 times.

Also, after about an hour or so, it seems like it looses functionality. I can still crawl through my passes, but pressing ENTER on them or clicking and using *any* function (fill, goto and fill) doesn't do anything. I have to de- and re-activate the plugin (enter passwords 3 times...) and then it works again.

What did I fuck up?

In my terminal, `pass` works wonderful and without problems, so I guess that's not the problem.

This is my config: [http://i.imgur.com/DWqSKPI.png](http://i.imgur.com/DWqSKPI.png)
Plugin version is 1.0.0

If you need any other information, pls don't hesitate to ask. :-)

drmoose:
This PR:
* Keeps a map of running execPass calls to prevent starving the gpg-agent with multiple overlapping requests for the same password. On my system, this dramatically reduces the number of duplicate passphrase prompts I receive.
* Suppresses the usual `window.alert` on `pass` failure when called from `getPasswordData`, which would fight with the gpg-agent passphrase dialog for focus and cause the UI to become unresponsive.
* Removes the infinite loop that caused passff to freeze the browser when the `No secret key` error occurs.

Tested against FF Nightly 59.0a1 (2017-11-19) (64-bit) on ArchLinux.

It is believed this PR will close #220, #243 and #101.

### DIFF
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -131,7 +131,7 @@ PassFF.Pass = {
 
     if (item.isLeaf()) { // multiline-style item
       let args = [item.fullKey()];
-      return this.executePass(args).then((executionResult) => {
+      return this.executePass(args, {}, true).then((executionResult) => {
       if (executionResult.exitCode !== 0) {
         return;
       }
@@ -427,7 +427,7 @@ PassFF.Pass = {
     });
   },
 
-  executePass: function(args, subprocessOverrides) {
+  executePass: function(args, subprocessOverrides, blockFailAlert) {
     let result = null;
     let scriptArgs = [];
     let command = null;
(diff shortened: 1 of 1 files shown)


tuxor1337:
Thanks again!

Issue #182, opened 2017-01-16
Filling a form doesn't generate events


toupeira:
I ran into an issue with Grafana, when I fill the login form with PassFF the login button stays disabled because it's apparently listening for key events ([similar issue in older Grafana version here](https://github.com/grafana/grafana/issues/3055)). I didn't encounter this issue with LastPass so I decided to open a bug report here, instead of reopening the Grafana issue :)

If I trigger a `change` event on the input fields (using `$('input').change()` in the browser console) the login button becomes active, maybe you could add this in PassFF after a form fill?

tuxor1337:
Nice idea, but we don't have jQuery and using plain JavaScript (`.dispatchEvent(new Event("change"))`, same with "keyup" and "keydown"), I don't get the expected result since Grafana is listening to other events. Any other suggestions?

grodzik:
How about something like this?:

```
diff --git a/src/modules/page.js b/src/modules/page.js
index 6ee488c..9eda4a7 100644
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -55,20 +55,37 @@ PassFF.Page = {
   },

   setInputs: function(doc, passwordData) {
     PassFF.Page.setLoginInputs(doc, passwordData.login);
     PassFF.Page.setPasswordInputs(doc, passwordData.password);
     PassFF.Page.setOtherInputs(doc, passwordData._other);
   },

+  fireOnChange: function(doc, element) {
+    var evt = doc.createEvent("HTMLEvents");
+    evt.initEvent("change", false, true);
+    element.dispatchEvent(evt);
   },

   setLoginInputs: function(doc, login) {
     PassFF.Page.getLoginInputs(doc).forEach(function(loginInput) {
       loginInput.value = login;
+      PassFF.Page.fireOnChange(doc, loginInput);
     });
   },

   setPasswordInputs: function(doc, password) {
     PassFF.Page.getPasswordInputs(doc).forEach(function(passwordInput) {
       passwordInput.value = password;
+      PassFF.Page.fireOnChange(doc, passwordInput);
     });
   },

@@ -82,6 +99,7 @@ PassFF.Page = {
       }
       if (value) {
         otherInput.value = value;
+        PassFF.Page.fireOnChange(doc, otherInput);
       }
     });
   },
```

[...]

PR #249, opened 2017-11-19
Generate DOM events during form fill, closing #182.

drmoose:
The previous PR, #232 by @GordanV addressed #182 by firing an "input" event after filling each field. @tuxor1337 said "Thanks!" and then closed the PR without further explanation. The event-generating logic never made it into master. This is an attempt at reviving that effort.

Loosely based on Gordan's work, this PR tries to simulate more of the DOM events that occur when one interacts with an inputbox, albeit, only once (rather than once per character).
* `keydown`
* `keyup`
* `keypress`
* `input`
* `change`

Tested against FF Nightly 59.0a1 (2017-11-19) (64-bit) on ArchLinux, on these sites:
* https://www.hover.com/signin (from [here](https://github.com/passff/passff/issues/182#issuecomment-333990039))
* https://verified.capitalone.com/sic-ui/#/esignin?Product=Card (from [here](https://github.com/passff/passff/issues/182#issuecomment-313843330))
* http://play.grafana.org/login (from [here](https://github.com/passff/passff/issues/182#issuecomment-310819637))
* https://www.netflix.com/login (a thorn in my side since I started using password managers)


tuxor1337:
Wow, why didn't anybody complain before? That was a mistake, I didn't want to discard @GordanV's PR! To give @GordanV his due credit, I merged his PR now (it also cover's some submit button logic, so that makes sense anyways).

Thanks a lot for your work @drmoose! It indeed closes #182, finally coping with the more nasty cases mentioned there.

PR #250, opened 2017-11-20
Add some DOM trickery to let the keyboard shortcut through, even when a password entry field is focused.

### LINKED ISSUE #225: Shortcut does not activate plugin while focused on textbox
Control-Y does not activate the plugin if I'm focused a textbox (let's say username), it DOES work when I unfocuse the textbox by clicking elsewhere. I'm on Arch Linux with Firefox Developer Edition and unsigned 1.0.0 release.

drmoose:
This PR is intended to close #225. 

As requested in that issue, this commit attempts future-proofing against the eventual solution to [the configurable keyboard shortcuts Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1215061). 

The central idea is to blur any focused text entry fields on `keydown` so they can't swallow the `keypress` that would otherwise trigger the passff UI. Most of the logic involved is in calculating what we're expecting the `keydown` event to look like based the browser.commands configuration. A smaller hardcoded version might look like this:

```javascript
document.addEventListener('keydown', (evt) => {
  if ('y' == evt.key && evt.ctrlKey)
    document.firstElementChild.focus();
});
```

Tested against FF Nightly 59.0a1 (2017-11-19) (64-bit) on ArchLinux.

### DIFF
--- a/src/modules/main.js
+++ b/src/modules/main.js
@@ -137,6 +137,67 @@ var PassFF = {
     let items = PassFF.Pass.getUrlMatchingItems(PassFF.tab_url);
     PassFF.menu_state.items = items.map((i) => { return i.toObject(true); });
     PassFF.Page.tabAutoFill(tab);
+
+    // Allow our browser command to bypass the usual dom event mapping, so that
+    // Ctrl+Y still works, even  when a password field is focused.
+    browser.commands.getAll().then((commands) => {
+      // Mapping between modifier names in manifest.json and DOM KeyboardEvent.
+      let commandModifiers = {
+        'Ctrl': browser.runtime.PlatformOs == 'mac' ? 'Meta' : 'Control',
+        'MacCtrl': 'Cotrol',
+        'Command': 'Meta',
+        'Alt': 'Alt',
+        'Shift': 'Shift'
+      };
+
+      // Read the _execute_browser_action command to get its shortcut. We're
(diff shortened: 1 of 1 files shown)


tuxor1337:
Looks very good to me. It's probably the best we can do at the moment (without having the upstream bug fixed).

drmoose:
Agreed this one definitely needs thorough review. In particular:

* Is `document_idle` the right value for [`runAt`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/extensionTypes/RunAt)? Leaving it blank was a guess that seemed to work on the sites I was testing.
* Can somebody with a Mac verify I've got the control shortcut mapping logic right? It's based entirely on the description on MDN.
* Is there a performance hit implication doing this setup in tab_init, which is called many times on the same tab? (This is also the reason the auto-filler is so spammy, but I haven't had time yet to work out why it's happening in the first place.) 

I meant to include these notes in the original description, but got distracted trying (unsuccessfully) to get a macOS test environment rigged up. Sorry about that.

tuxor1337:
The Mac question will pop up again anyway once we have found somebody with a Mac. Also, the trouble about tab_init has to be reviewed later due to the auto-filler problems. So I merged your proposals (after some further restructuring) just to have some activity here, at last ;-)

PR #254, opened 2017-12-08
Copy to clipboard in page script, fix #233

### LINKED ISSUE #233: "Copy login" / "Copy password" do not work when GPG requires password
I am on: Firefox 55.0.3 x64 on Fedora 26 running KDE, running pass 1.7.1 and passff 1.0.3linux.

When selecting "Copy login" or "Copy password" from the passff menu, if the GPG key must be unlocked, the requested field is not copied to the clipboard after completing unlocking. It works correctly if unlocking is not required.

To reproduce:

1. Restart gpg-agent so nothing is in the keyring
2. Go to the passff menu and select "Copy login" or "Copy password" for any entry
3. gpg-agent should open a window asking for the password for your GPG key
4. Enter your password, adding the key to the keyring
5. Attempt to paste your login/password from the clipboard
6. Observe that this works as expected after trying again, once the key is in the keyring. Also, notice that repeating these steps with "Fill" instead of copy successfully fills the information after unlocking the keyring; the issue is limited only to copying

Expected behavior: the login/password should be copied into the clipboard

tuxor1337:
Fixes #233 by adapting the copy to clipboard code from https://github.com/mdn/webextensions-examples/tree/master/context-menu-copy-link-with-types

The previous code did the copying inside the browser action menu which had to remain open until the password data was received from the password store. In cases where the GPG agent would try to open a password prompt, the UI would run into a focus race condition between the password prompt and the browser action menu. In some situations this would even cause Firefox to crash.

### DIFF
--- a/src/modules/clipboard-helper.js
+++ b/src/modules/clipboard-helper.js
@@ -1,0 +1,20 @@
+// Code taken from `context-menu-copy-link-with-types/clipboard-helper.js`
+// as included in repository https://github.com/mdn/webextensions-examples/
+
+// This function must be called in a visible page, such as a browserAction popup
+// or a content script. Calling it in a background page has no effect!
+function copyToClipboard(text, html) {
+    function oncopy(event) {
+        document.removeEventListener("copy", oncopy, true);
+        // Hide the event from the page to prevent tampering.
+        event.stopImmediatePropagation();
+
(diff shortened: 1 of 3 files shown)


tuxor1337:
This is obsolete by commit c17fb4d891bcba5580b98588ae5184e849afa100

PR #267, opened 2018-01-07
Fix platform detection for macOS

vizv:
This fix the issue that `Pass command` and `Environment variables` field didn't set as expected in macOS.

Note: `browser.runtime.PlatformOs` returns an object instead of a string, and `getPlatformInfo()` is the correct method to be used (you might want to check in the console of an add-on).

For more information, check here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/runtime/getPlatformInfo

Btw, I didn't change another `browser.runtime.PlatformOs` instance since the `command` + `y` shortcut works in macOS for some reason, and I prefer not to touch it.

### DIFF
--- a/src/modules/preferences.js
+++ b/src/modules/preferences.js
@@ -94,10 +94,12 @@ PassFF.Preferences = (function () {
     'autoFillBlacklist'
   ];
 
-  if (browser.runtime.PlatformOs === "mac") {
-    prefParams.command = '/usr/local/bin/pass';
-    prefParams.commandEnv = 'PATH=/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin';
-  }
+  browser.runtime.getPlatformInfo().then((info) => {
+    if (info.os === "mac") {
+      prefParams.command = '/usr/local/bin/pass';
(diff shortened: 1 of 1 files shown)


vizv:
Oh, I forgot to mention.

`/usr/sbin` and `/sbin` is not required since no program under those `PATH` was used.
`/usr/local/bin` might also not required, since `pass` is the only program needed in this `PATH` but I didn't verify that, if you could wait me for a day I'll verify this.

Another two directories `/usr/bin` and `/bin` are mandatory since:
* `getopt` in `/usr/bin` are required, without it, `pass` script will run in the loop (check function `show`, there is a `while true` with a `case` statement, when getopt fails, an empty `$1` will result `pass` script running in a busy loop)
* lots of other programs in `/bin` are also required.

tuxor1337:
Thanks a lot! Since I don't have access to a Mac machine, it would be good to have somebody else check this... anybody?

By the way, I find the documentation of `browser.runtime.PlatformOs` quite misleading...

tuxor1337:
These do not apply any more since all the relevant preferences have moved to the host app script.

Issue #266, opened 2018-01-06
OSX: `sed` syntax in host install script


tuxor1337:
User @mddvul22 reported in https://github.com/passff/passff/issues/219#issuecomment-355705850 about problems when executing the `install_host_app.sh` script on his Mac machine. The output is:

    $ curl -sSL https://github.com/passff/passff/releases/download/1.1/install_host_app.sh | bash -s -- firefox
    Python 3 executable located at /usr/local/bin/python3
    Installing Firefox host config
    sed: 1: "/Volumes/Data/Users/cib ...": extra characters at the end of D command

My first guess is that it might have to do with the difference between `sed` (BSD style) and `gsed` (GNU style). But I'm not entirely sure. We need somebody with Mac to do some testing here, because it's not clear which of the two sed commands fails or whether it can be solved by replacing sed with gsed etc.

vizv:
I'll look into this, since I installed GNU flavoured `sed` with `brew` and I forgot to mention that.

jicksta:
I'm getting the following on High Sierra:

```
$ ./install_host_app.sh --local firefox
Python 3 executable located at /usr/local/bin/python3
Installing Firefox host config
sed: 1: "/Users/jay/Library/Appl ...": invalid command code j
```

tuxor1337:
@jicksta Thanks! Have you tested with vizv's proposed changes from here: https://github.com/vizv/passff/blob/f5cd7a30d3e1214c1ffc89bba7e8b87432417a39/src/host/install_host_app.sh

jicksta:
@tuxor1337: @vizv's script does execute without any errors (see below), but unfortunately my extension still seems unable to communicate with `pass`.

```
❯ ./install_host_app.sh --local firefox
Python 3 executable located at /usr/local/bin/python3
Installing Firefox host config
Native messaging host for Firefox has been installed to /Users/jay/Library/Application Support/Mozilla/NativeMessagingHosts.
```

Perhaps the extension could warn when the NativeMessagingHost is not installed or it's having trouble getting an index of entries from the host system?

vizv:
@jicksta did you remove the `NativeMessagingHosts` before install the new script?

When installing with `--local` option, did you place the required files `passff.json` and `passff.py` in the same directory?

After installation is completed, could you verify shebang is replaced correctly for `~/Library/Application Support/Mozilla/NativeMessagingHosts/passff.py`?

One last thing, you might need to disable / re-enable passff in add-on page.

[...]

PR #268, opened 2018-01-07
Fix syntax error using BSD flavored sed for macOS

vizv:
This PR uses BSD flavored `sed` comes with macOS, and the updated script has been tested on macOS High Sierra 10.13.2 on my laptop, which should fix issue #266.


wezm:
Probably should use the BSD sed style on BSD as well as macOS.

tuxor1337:
Would you please add this as a pull request for the new `passff-host` repository? https://github.com/passff/passff-host/pulls

PR #281, opened 2018-02-02
disable autocompletion while popup is open

### LINKED ISSUE #278: "fillable input fields icon": password list appears behind firefox' autofill
Sorry for the incomprehensible title, but I don't know how else to describe this:      
screenshot_2018-02-01_14-10-18

Is this something that PassFF can fix?

on a side note: this form is at `https://lms.uibk.ac.at` --  PassFF is also matching a whole lot of other, unrelated stuff (**E**: the one I need is the first and hidden beneath the autofill)

girst:
fixes #278 

I haven't tested this too much, since I'm not really a javascript developer, but it *works on my machine!™*

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -204,7 +204,8 @@ PassFF.Page = (function () {
        * overlap passff's popup menu. Also save its value beforehand, so it can
        * be restored when the popup gets dismissed.
        */
-      e.target.setAttribute("passff-autocomplete", e.target.autocomplete);
+      if (e.target.hasAttribute('passff-autocomplete'))
+        e.target.setAttribute("passff-autocomplete", e.target.autocomplete);
       e.target.autocomplete="off";
 
       return;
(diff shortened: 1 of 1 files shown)


tuxor1337:
I don't know whether there are users of PassFF out there that do use Firefox's autofill. In that case, PassFF will break this feature for them. Have you tried to apply your magic not on mouseover but on click? Of course, you would have to make sure that autocomplete is reset to the original value when the popup is closed.

girst:
>PassFF will break this feature

it won't. the patch disables autofill only when you are hovering above the (p) button and re-enables it when you move your mouse away. 

>Have you tried to apply your magic not on mouseover but on click

yes, at that time the autofill is already open and we're "too late" 

>you would have to make sure that autocomplete is reset

that is implemented, see line 214

girst:
you were right -- i had a bug in the code. autocompletion-reenabling is now working. 

onmouseover was firing multiple times, overwriting the saved value. i now only write to the save-state-attribute if it doesn't exist.

tuxor1337:
Okay, thanks!

PR #287, opened 2018-02-18
Fix #280

### LINKED ISSUE #280: Filling other fields doesn't seem to work
I've never been able to get passff to fill "lines besides the login and URL", as described in the documentation:

> Lines besides the login and URL that match the format `<other_inputfield_name>: <value>` can be used to fill in input fields besides the login and password fields. The left hand side of the colon should match the input field's name or id attribute.

Take for instance this form: https://www.qantas.com/fflyer/dyns/LoginForm

I now created a file `quantas.com.au/martin` with the following content:

    1234
    url: https://www.qantas.com/fflyer/dyns/LoginForm
    login: 1234567890
    memberId: 1234567890
    memberSurname: Jones

but it doesn't work at all, and only the PIN field is filled, while the login/memberId gets put in the newsletter field further down.

Any clues?

5bentz:
Please look at the commit `Version 1.1`. Is it correct?

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -158,10 +158,12 @@ PassFF.Page = (function () {
   function setOtherInputs(other) {
     getOtherInputs(other).forEach(function (otherInput) {
       let value;
-      if (other.hasOwnProperty(otherInput.name)) {
-        value = other[otherInput.name];
-      } else if (other.hasOwnProperty(otherInput.id)) {
-        value = other[otherInput.id];
+      let name = (otherInput.name).toLowerCase();
+      let id = (otherInput.name).toLowerCase();
+      if (other.hasOwnProperty(name)) {
+        value = other[name];
+      } else if (other.hasOwnProperty(id)) {
+        value = other[id];
       }
       if (value) {
         writeValueWithEvents(otherInput, value);


tuxor1337:
Thanks for the fix of #280. However, I don't think it's very elegant to hardcode the version number. Instead, I suggest to stay with the dummy testing placeholder and build a new "official" version for AMO using `make VERSION=1.5.3` etc.

5bentz:
I removed the version fix.

I agree that it would be better to build at the same location if it fixes the weird "XPI file is corrupted".
It also simplify the release process for you.

Basically, if I understand:
a dummy version name as default
```
Makefile:VERSION := testing
```
make replaces a placeholder with the right version
```
Makefile:       sed -i "s/_VERSIONHOLDER_/$(VERSION)/g" $@
Makefile:       sed -i "s/_VERSIONHOLDER_/$(VERSION)/g" $@
```
occurences of the placeholder
```
host/install_host_app.bat:SET "VERSION=_VERSIONHOLDER_"
host/install_host_app.sh:VERSION="_VERSIONHOLDER_"
manifest.json:    "version": "_VERSIONHOLDER_",
```

5bentz:
@tuxor1337 Thanks!
Btw, why convert field names ("login", "url", "memberId", etc) to lower case? 
Answer if you are knowledgeable about that ofc ;)

tuxor1337:
Some users like to write in lower-case:

    super-secret-pw
    login: foo
    url: bar

others prefer something like this:

    super-secret-pw
    Login: foo
    URL: bar

So, first thing we do when parsing the password files is lower-casing all field names - just for the dev's convenience. So far, this didn't have any side-effects ;-)

Issue #294, opened 2018-02-21
Input field's (P) menu can be offset


5bentz:
#### Describe your Environment
**Firefox:** dev & ESR
**passff:** AMO version & last version

#### Describe *objectively* your problem. What is happening?

On any webpage, resize the window in such a way to have an horizontal scrollbar.
Then, the input (P) menu is offset.

For example
* https://www.qantas.com/fflyer/dyns/LoginForm (easy to do)
* https://lwn.net/login (harder)

![bug_passff_input_context_menu_offset](https://user-images.githubusercontent.com/10729484/36482074-53ee66ee-1712-11e8-9a54-8aabf7629c99.png)

#### What should happen? What have you expected to happen?
It shouldn't be offset!

Anecdotal? Since most websites have responsive designs and the offset is rarely an issue (ie. when the menu goes *off-screen*).

PR #295, opened 2018-02-21
WIP: Issue280 rewrite

5bentz:
Here is the big PR, although the net change is somewhat small.

**This is a WIP** since it changes drastically the behaviour of passff. See 45655eb
**Do not merge**

I will comment commit per commit:

#### 3c0ba5e Refactor function setOtherInputs()

After fixing the fix for issue #280, I noticed the function was wrong: it finds inputs which _ids, names or autocompletes_ match the other fields. Then, it finds which field matches a given input by comparing only _ids and names_.
Thus, I have rewritten setOtherInputs() by using readInputNames().

From the commit message
```
Modified behavior:
* use readInputNames(): check 'autocomplete' attribute besides name & id
* if more than one field matches, use id or 'autocomplete' or name (in this order)
  instead of using name or id.
```

#### 71d7da7 readInputNames() sanitizes its output
It could return _undefined_ object for the autocomplete field. 
This commit also renders useless the check in isGoodName().
I have also split the array declaration on multiple lines for readability and for limiting the line to 80/90 chars.
Does it follow coding standards?  
Feel free to comment and propose another writing. 

Commit message:
```
* filters false values: NaN, undefined, null
* convert to lowercase
```

#### 45655eb Remove isGoodName() & refactor hasGoodName()
Kidding, it renders useless the whole function isGoodName().
*Actually*, this commit is problematic. Since passff adds fewer (P) icon on input fields.

(half an hour later)
Oh, isGoodName(name, goodNames) does not check if _name_ **IS** a _goodName_, it checks if _name_ **CONTAINS** a _goodName_. 
I wasn't aware of this. I'm pretty sure this is correct behaviour for you ;)

Thus, setOtherInputs() was all the more wrong since it checks if a _name_ **is** a _goodName_, wasn't it? :sweat_smile: 

Finally, I guess isGoodName() shouldn't be removed. 
However, could we rename it to containsGoodName? or even a more generic signature contains(container, elt).
hasGoodName() could be renamed to matchesGoodName() or matchesName() or hasMatchingName() (see `man grep` for example)
findGoodFieldNames -> findMatchingFieldNames()
Let me know what you think about these changes.

#### 37ef803 Change hasGoodName() signature. Refactor parents
This change makes sense in 3 ways:
* it prepares the next commits
* all its calls use readInputName() as argument fieldNames. 
* the signature hasGoodName(input, goodNames) makes more (semantic) sense: Does input has good names?

#### bb759bd Add findGoodFieldNames(). Refactor setOtherInputs()
setOtherInputs() does too many things. Let's hand down some lines to a new function.
The new function is basically hasGoodName() with a different output: instead of returning yes/no, it returns an array of matching elements.

#### 5861221 Refactor hasGoodName()
It's free.
Note: This last commit was breaking the field matching. There were no (P) icons anymore and passff couldn't fill any field anymore. But a rebase on the master branch fixed it.

I found a bug during the test process: I expose it there since it can be reproduced with my bug (see 45655eb)
* `Add login input name` in input's contextual menu not working for GitHub's sign up fields https://github.com/join . It fails **silently** to add the name "user[login]". Yes this is silly. We could also add an option to add the id.

Another anecdotal bug https://github.com/passff/passff/issues/294

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -156,17 +156,14 @@ PassFF.Page = (function () {
   }
 
   function setOtherInputs(other) {
+    const goodFieldNames = Object.keys(other);
     getOtherInputs(other).forEach(function (otherInput) {
-      let value;
-      let name = (otherInput.name).toLowerCase();
-      let id = (otherInput.id).toLowerCase();
-      if (other.hasOwnProperty(name)) {
-        value = other[name];
-      } else if (other.hasOwnProperty(id)) {
-        value = other[id];
-      }
-      if (value) {
-        writeValueWithEvents(otherInput, value);
+      let inputFieldNames = readInputNames(otherInput);
+      inputFieldNames = inputFieldNames.filter(Boolean).map((n) => n.toLowerCase());
+      inputFieldNames = inputFieldNames.filter((n) => goodFieldNames.indexOf(n) >= 0);
+      const writtenField = inputFieldNames.pop();
+      if(writtenField) {
+        writeValueWithEvents(otherInput, other[writtenField]);
       }
     });
   }
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -58,6 +58,12 @@ PassFF.Page = (function () {
     ].filter(Boolean).map((n) => n.toLowerCase());
   }
 
+  function findGoodFieldNames(input, goodFieldNames) {
+    const fieldNames = readInputNames(input);
+    goodFieldNames = goodFieldNames.filter(Boolean).map((n) => n.toLowerCase());
+    return fieldNames.filter((n) => goodFieldNames.indexOf(n) >= 0);
+  }
+
   function hasGoodName(input, goodFieldNames) {
     const fieldNames = readInputNames(input);
     goodFieldNames = goodFieldNames.filter(Boolean).map((n) => n.toLowerCase());
@@ -158,9 +164,7 @@ PassFF.Page = (function () {
   function setOtherInputs(other) {
     const goodFieldNames = Object.keys(other);
     getOtherInputs(other).forEach(function (otherInput) {
-      let inputFieldNames = readInputNames(otherInput);
-      inputFieldNames = inputFieldNames.filter((n) => goodFieldNames.indexOf(n) >= 0);
-      const writtenField = inputFieldNames.pop();
+      const writtenField = findGoodFieldNames(otherInput, goodFieldNames).pop();
       if(writtenField) {
         writeValueWithEvents(otherInput, other[writtenField]);
       }
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -58,12 +58,8 @@ PassFF.Page = (function () {
     ].filter(Boolean).map((n) => n.toLowerCase());
   }
 
-  function isGoodName(name, goodNames) {
-    return goodNames.some((n) => { return name.indexOf(n.toLowerCase()) >= 0; });
-  }
-
   function hasGoodName(fieldNames, goodFieldNames) {
-    return fieldNames.some((fn) => { return isGoodName(fn, goodFieldNames); });
+    return goodFieldNames.some((n) => fieldNames.indexOf(n.toLowerCase()) >= 0);
   }
 
   function isPasswordInput(input) {
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -58,27 +58,29 @@ PassFF.Page = (function () {
     ].filter(Boolean).map((n) => n.toLowerCase());
   }
 
-  function hasGoodName(fieldNames, goodFieldNames) {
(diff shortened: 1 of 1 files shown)


tuxor1337:
@5bentz Thanks for this PR! I made some adaptations and hope to have addressed all your concerns in the 08cbffd

tuxor1337:
Concerning your comment on adding `user[login]` as a new input name: Since this string matches a lookup for `login` (which is one of the default input names), it is not added on purpose!

5bentz:
Thanks for the merge!

My concerns were code correctness & DRY principe ;) I read the code and it looks fine to me. I especially like the `findMatchingName` function, even if its style is different from mine. It clearly shows its genericity. Is `findMatchingName` a function you have found somewhere?
I notice that `hn.indexOf` shouldn't fail (TypeError: hn is null) since readInputNames filters the truthness of hn :+1: 

On a side note: the README is accurate: it claims passff searches strings in the HTML page **containing** the specified values, which is true! :+1:  
> From the extension preferences you will be able to set:
> Inputs (A comma separated list of input names. Input field names in a html page **containing** one of those values will be filled with the corresponding value.)

tuxor1337:
Thanks for your approval! The function `findMatchingName` is the result of your proposal to write a function `findMatchingFieldNames`. Maybe it would be slightly more generic defining a function

    function containsName(name, haystack, exact) {
        // check if haystack contains name
        exact = exact || false;
        return haystack.some(hn => {
            return (!exact && hn.indexOf(name.toLowerCase()) >= 0
                        || (hn === name.toLowerCase()));
        });
    }
    
    function findMatchingName(needles, haystack, exact) {
        // return first needle that is found in haystack
        return needles.find(name => containsName(name, haystack, exact));
    }

This is slower than the current version if the array `haystack` is significantly larger than `needles`. In our case, the `haystack` is at most as large as the number of input fields on the page, which is usually small. So we could just as well use this version...

5bentz:
What do you mean by *more generic*?
The latter version of `findMatchingName` does exactly the same as the current version in  https://github.com/passff/passff/commit/08cbffd36f4c741046dc675f0fc3e62c4517f7be

Why should it be slower?
Except the different traversal (for each needle, traverse the haystack VS haystack.forEach(traverse the needles)), I cannot see any difference. 
However, it *may* be faster because it starts by checking the first needles, which may be more likely to match the haystack. For example, login input names "login", "mail".
Finally, one (me) may say such optimisations are irrelevant without a proper flame graph analysis. 

I like the idea of refactoring but I don't see the point of your proposition. Are you resurrecting `hasGoodName`?
I believe it can make sense to recreate a helper function for `findMatchingName` if it makes the code more readable or if this helper function is useful somewhere else in the code.


Your code gave me an idea of a more generic function: `exact` could be replaced with the actual test to perform on each pair of elements.

Signature: `find(haystack, needles, test)` with `test` being a test function of signature `test(hn, n)` (notice that its signature keeps the same order: haystack, needle).

The code:

```
function find(haystack, needles, test) {
    for(let n of needles) {
        let foundElt = haystack.find(hn => test(hn, n));
        if(foundElt) return foundElt;
    }
    return undefined
}

# for people who do not like two returns in the same function:

function find(haystack, needles, test) {
    let foundElt = undefined;
    let i = 0;
    while(foundElt === undefined && i < needles.length) {
        let n = needles[i];
        foundElt = haystack.find(hn => test(hn, n));
    }
    return foundElt;
}
```
In our situtation, `test` would be
```
(hn, n) => hn.indexOf(n.toLowerCase()) >= 0
or 
(hn, n) => hn === n.toLowerCase())
```

Pros +
+ is a fully generic function: we can test anything with any pair of arrays
+ uses `Array.find`
+ behaves like `Array.find` (if nothing is found, returns `undefined`)

Cons -
- I don't know

tuxor1337:
Okay, let's not discuss about which version is slower. This is a priori not clear at all, as you pointed out correctly.

However, note that we want to return the matching needle and not the matching element from the haystack (as in your proposal). Still, I like your idea of introducing a `test` callback. The situation is almost symmetric, because we are looking for an element in the intersection of two arrays. The asymmetry is introduced by the `test` function. How about this:

    function findIntersection(arr1, arr2, callback) {
        // find first element from arr1 in intersection of arr1 and arr2
        // equality of elements is determined according to callback(el1, el2)
        return arr1.find(el1 => arr2.some(el2 => callback(el1, el2)));
    }
    
    findIntersection(goodNames, inputFieldNames, (hn, n) => hn.indexOf(n.toLowerCase()) >= 0);
    findIntersection(goodNames, inputFieldNames, (hn, n) => hn === n.toLowerCase());

According to DRY we should probably give the function `(hn, n) => hn.indexOf(n.toLowerCase()) >= 0` a name because it's used in two places.

PR #318, opened 2018-03-30
[Discussion] Improve English translation

5bentz:
Source: French translation (See PR #319  :+1: )

#### Main points

- Use a vocabulary closer to the user experience, and farther from the code, for eg:
  - input | input fields | form > form (*field* is already used for the password data)
  - root > all
  - overwrite > replace
- Improve/Correct
  - url > URL
  - enter key > Enter key
  - URL > tab's URL (more precise)
  - *menu > *toolbar menu* or *form menu*
  - symbols > special characters

#### What could be improved?
- Maybe punctuations. 
- I don't know, tell me!

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -18,7 +18,7 @@
     "message": "Fill and submit"
   },
   "passff_menu_goto_fill": {
-    "message": "Goto, fill"
+    "message": "Goto and fill"
   },
   "passff_menu_goto_fill_and_submit": {
     "message": "Goto, fill and submit"
@@ -30,7 +30,7 @@
     "message": "Display"
   },
   "passff_display_hover": {
-    "message": "Hover mouse pointer to show password"
+    "message": "Hover mouse pointer to show password."
   },
   "passff_toolbar_button_label": {
     "message": "PassFF"
@@ -48,7 +48,7 @@
     "message": "New"
   },
   "passff_button_root_label": {
-    "message": "Display all"
+    "message": "Display top-level"
   },
   "passff_button_context_label": {
     "message": "Display contextual"
@@ -75,10 +75,10 @@
     "message": "Miscellaneous"
   },
   "passff_prefs_mark_fillable": {
-    "message": "Mark fillable forms with an icon-shaped form menu"
+    "message": "Mark fillable input fields with the PassFF icon"
   },
   "passff_prefs_submit_fillable": {
-    "message": "Submit when selected via the form menu"
+    "message": "Submit when selected via the input menu"
   },
   "passff_prefs_handle_http_auth": {
     "message": "Handle HTTP authentication"
@@ -93,13 +93,13 @@
     "message": "Behavior of Enter key"
   },
   "passff_prefs_fields_description": {
-    "message": "Fields are used to find data inside your password repository. PassFF looks login, password and URL up in the password file. If it's a directory, PassFF looks them up in the files of this directory"
+    "message": "Fields are used to find data inside your password repository. PassFF looks login, password and URL up in the password file. If it's a directory, PassFF looks them up in the files of this directory."
   },
   "passff_prefs_password_input_names_label": {
-    "message": "Password form names"
+    "message": "Password input names"
   },
   "passff_prefs_login_input_names_label": {
-    "message": "Login form names"
+    "message": "Login input names"
   },
   "passff_prefs_login_field_names_label": {
     "message": "Login field names"
@@ -111,7 +111,7 @@
     "message": "URL field names"
   },
   "passff_prefs_script_howto": {
-    "message": "See passff.py for changing the environment variables of Pass!"
+    "message": "See passff.py for changing the environment variables of pass."
   },
   "passff_prefs_auto_fill_label": {
     "message": "Try to auto fill page login forms"
@@ -123,7 +123,7 @@
     "message": "Don't auto fill if tab's URL contains one of the strings"
   },
   "passff_prefs_default_length_tooltip": {
-    "message": "The default value for the length of a generated password"
+    "message": "Default length of a generated password"
   },
   "passff_prefs_default_length_label": {
     "message": "Default password length"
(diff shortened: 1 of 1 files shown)


src/_locales/en/messages.json
```diff
@@ -48,7 +48,7 @@
     "message": "New"
   },
   "passff_button_root_label": {
-    "message": "Display root"
+    "message": "Display all"
```
tuxor1337: I don't agree with this: Displaying "all" password entries is different from displaying the "root" (directory) of the password store.

src/_locales/en/messages.json
```diff
@@ -75,31 +75,31 @@
     "message": "Miscellaneous"
   },
   "passff_prefs_mark_fillable": {
-    "message": "Mark fillable input fields with icon"
+    "message": "Mark fillable forms with an icon-shaped form menu"
```
tuxor1337: Even though I see the confusion that the word `field` is already used to describe the fields in password entries, I don't think it's correct to say we are marking the forms with icons. Furthermore, it's not the menu that is icon-shaped.

src/_locales/en/messages.json
```diff
@@ -75,31 +75,31 @@
...
-    "message": "Behavior of enter key"
+    "message": "Behavior of Enter key"
   },
   "passff_prefs_fields_description": {
-    "message": "Fields are used to find data inside your passwords repository. PassFF will try to find the login, password or url value inside the password data (using the standard syntax) or trying to find a matching child node name"
+    "message": "Fields are used to find data inside your password repository. PassFF looks login, password and URL up in the password file. If it's a directory, PassFF looks them up in the files of this directory"
   },
   "passff_prefs_password_input_names_label": {
-    "message": "Password input names"
+    "message": "Password form names"
```
tuxor1337: What we are refering to here, is not a form, it's only an input field in a form.

src/_locales/en/messages.json
```diff
@@ -75,31 +75,31 @@
...
   },
   "passff_prefs_login_input_names_label": {
-    "message": "Login input names"
+    "message": "Login form names"
```
tuxor1337: Same as above.

src/_locales/en/messages.json
```diff
@@ -108,10 +108,10 @@
     "message": "Password field names"
   },
   "passff_prefs_url_field_names_label": {
-    "message": "Url field names"
+    "message": "URL field names"
   },
   "passff_prefs_script_howto": {
-    "message": "See passff.py for changing the command environment!"
+    "message": "See passff.py for changing the environment variables of Pass!"
```
tuxor1337: I'm not sure, whether it's comprehensible to capitalize the name of zx2c4 pass.

src/_locales/en/messages.json
```diff
@@ -132,10 +132,10 @@
     "message": "Whether generated passwords should include symbols by default"
   },
   "passff_prefs_default_include_symbols_label": {
-    "message": "Include symbols by default"
+    "message": "Include at least one special character in the password by default"
   },
   "passff_prefs_preferred_new_method_tooltip": {
-    "message": "Whether new passwords default to being generated or being inserted"
+    "message": "Whether new passwords default to be automatically generated or to be inserted by user"
```
tuxor1337: The verb "to default to sth." is used with the "-ing" form of the verb (source: my experience + Google). Can you give a source for the contrary?

src/_locales/en/messages.json
```diff
@@ -147,13 +147,13 @@
...
-    "message": "Show 'New Password' button in menu"
+    "message": "Show 'New Password' button in the toolbar menu"
   },
   "passff_prefs_enable_logging": {
-    "message": "Enable logging to the Javascript console"
+    "message": "Enable logging to the JavaScript web console"
   },
   "passff_prefs_show_status": {
-    "message": "Show a status line in the toolbar menu"
+    "message": "Show a statusbar in the toolbar menu"
```
tuxor1337: I'd suggest "status bar" instead of "statusbar",

src/_locales/en/messages.json
5bentz: Oh yes I wasn't aware of this.
Displaying all passwords is rather a recursive list (think `ls -R`).
This button just displays the top-level directory.
Maybe `top-level`?
Example: Top-level Domain.

src/_locales/en/messages.json
5bentz: I'm not fond of the `icon-shaped form menu` either. Indeed, the menu doesn't have a shape of icon :+1:

src/_locales/en/messages.json
5bentz: Having fields for different entities is tough. Maybe there is a compromise: a [`form input`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input). But it complicates everything (2 words VS 1).

#### However, I have thought about a UX improvement that address this ambiguity

Put the *password/login input names* inputs in their own section called *Inputs* (or Form Inputs), like the Fields. And like the fields, add a text description such as:
`Input names are used by PassFF to find the fillable fields/forms/inputs in web pages.`
It separates the inputs names from the *Autofill* section, which is desirable since inputs names are also used to mark the inputs!

Preferences *Mark fillable inputs* and *Submit when selected via icon menu* could also be moved in this new section.

It might be the subject of a new PR.

src/_locales/en/messages.json
5bentz: Good point.
The title of the section `Pass Script` is capitalized since it is an English title I guess.

src/_locales/en/messages.json
5bentz: My bad, you are right.

5bentz:
From the commit message. (I will squash the commits anyway!)

#### Tuxor1337 correction have been applied
- I propose `Display top-level` instead of `root`. To be honest, I like both words.

#### Names of the two menus:
- *toolbar menu* seems perfect for a...menu in the toolbar!
- for the icon menu: I'd rather call it `input/form/field menu` instead of `icon menu`, insofar as the toolbar menu is an icon menu too!

#### Punctuation (The German translation has more punctuation marks!)
As a rule of thumb, I punctuate* when the sentence is told to the user (such as errors, questions, tips).
* This is my own rule. It makes sense to me. I don't know what the convention is. 

And a bench of other things & oversights.

#### Notes
- I kept a single *symbol* in the tooltip[1] to make sure ppl understand what we are talking about (If they happen not to know special characters...).
- About *Preferred new method. Generate. Insert*, I would like to be more specific but I failed to come up with a better translation. I edited the tooltip.

#### Off-topic
[1] How tooltips are triggered?

PR #333, opened 2018-06-01
Security Checks Before Filling Passwords

U8NWXD:
# Purpose
One security benefit of using password managers like `pass` is that, when they
are integrated with the browser through extensions like `passff`, they can make
phishing more difficult by comparing domains before filling passwords. Even if
the user is fooled into logging in to `examp1e.com` instead of `example.com`,
the extension can notice the difference and alert the user.

# `PassFF` Implementation
As of now, `PassFF` lacks this functionality and will fill any credentials the
user selects, even if domains do not match. This PR provides a very much
unpolished checking mechanism that warns the user in the following scenarios:

## When User is Warned
* If the domain of the current site and of the URL in the password database differ
  * Note that the current implementation does not correctly handle
  top-level-domains (TLDs) with multiple dot-separated strings like `.co.uk`.
  Because the TLD is assumed to be only one dot-separated string like `.com`,
  `www.example.co.uk` is interpreted as having a domain of `.co.uk` instead of
  the correct `example.co.uk`.
* If the current page URL is not served over HTTPS
  * This is again imperfect as the current implementation only checks the URL
  of the page, not the configuration of the login form itself, which could still
  be sending passwords in plaintext.
* If the URL in the password database is not HTTPS. This is a softer warning, as
there is no imminent security threat in this case, but keeping the URLs in the
database explicitly HTTPS helps protect the user in case they follow those links
to access websites that do not automatically redirect to secure pages.
* If there is an error parsing URLs. This can happen if `example.com` is stored
in the database instead of `https://www.example.com` as the URL constructor
expects. This simply serves to warn the user that anti-phishing protections will
be bypassed to avoid giving a false sense of security.

# Audience
The implementation in this PR is probably too strict for most users, so it can
be toggled from the extension's preferences and is off by default. This
preserves usability for a wider audience while accommodating users who want more
strict security settings. In addition, the necessary UI strings have not been
translated, making these changes of limited use to non-English speakers. As
such, I recognize that these changes may not be ready for deployment. However,
I would appreciate feedback and advice on how they might be improved.

# Still To Do
* Translate UI strings
* Parse incomplete domains like `google.com`
* Offer to automatically correct stored `http` links
* Make pop-up dialogs more aesthetic than the default `alert()` and `confirm()`
dialogs.
  * In the process, redesign the override confirmation dialogs to discourage
  overrides

# Potential Future Work
* Handle two-dot TLDs
* Prevent other password exfiltration attacks
  * Forms sending passwords insecurely
  * Forms with different destinations than before

# Attributions
* [Password Managers: Attacks and Defenses](https://www.usenix.org/conference/usenixsecurity14/technical-sessions/presentation/silver)
  * Authors:
    * Stanford University: David Silver, Suman Jana, and Dan Boneh
    * Carnegie Mellon University: Eric Chen and Collin Jackson
  * Publication: Proceedings of the 23rd USENIX Security Symposium
    * August 20–22, 2014
    * San Diego, CA
  * ISBN 978-1-931971-15-7


5bentz:
Thanks for the thorough PR! Well explained and I loved the reference (Password Managers: Attacks and Defenses).

#### About the PR (functional)

I understand this is a draft version of a domain check for PassFF, by watching the presentation on Password Managers, I noticed another **potential improvement** for this PR:

- Following the example of *Secure Filling*, we could check the domain the form sends to (step 4). The implementation seems not as straightforward as the current check though.

#### About the implementation
JS is new to me but I'll make a review anyway :+1:

src/modules/pass.js
```diff
@@ -154,6 +154,64 @@ PassFF.Pass = (function () {
...
+
+    if (passDomain != currDomain) {
+      return confirm( _("passff_domain_mismatch", [currDomain, passDomain]) +
+      "\n" + _("passff_override_antiphishing_confirmation"));
+    }
+
+    let passProt = passURL.protocol;
+    let currProt = currURL.protocol;
+
+    if (passProt != "https:") {
```
5bentz: This part below could be another function called `protocolSecurityWarning(passURL)`.
KISS :)
passItemURL -> passURL.origin

5bentz:
I have pushed a branch with the changes I mentioned above: https://github.com/5bentz/passff/tree/domain_check

1. Each check has its own function
2. The functions moved to page.js
3. They are called in fillActiveElements & fillInputs
4. Various improvements: use window.confirm, pass tab's URL.

It fixes the error "NS_ERROR_NOT_AVAILABLE" & the checks are passed as expected (by me).

I haven't touched the UI strings yet. They are quite unusual to me :o

U8NWXD:
@5bentz thank you for your review. I incorporated the changes you suggested, and I made the comparison at the end of `securityChecks` of the results of the helper functions more verbose. I think it is easier to understand the reasoning behind the comparisons now. I also tried to make the UI strings clearer, but could you clarify what is unusual about them? I also haven't looked into checking the form's destination domain yet.

5bentz:
@U8NWXD Thanks for the update :) It is indeed clearer than my cool one-liner :+1: 
I can load PassFF with your update, but it fails to do anything...strange. The code passes ESlint... I'll investigate that tomorrow.

> but could you clarify what is unusual about them? 

You insert variables in the strings such as `$passItemURL$` using a "placeholder" syntax, which is new to me.
```
    "placeholders": {
      "currDomain" : {
        "content" : "$1", // first argument
        "example" : "phishingDomain.com"
      },
      "passDomain": {
        "content" : "$2", // second argument
        "example": "realDomain.com"
      }
```
Until now,  we haven't declared any variable (read placeholder) in our UI Strings. That's why they are strange to me.

> This branch cannot be rebased due to conflicts

Keeping your feature branch up-to-date with the upstream branch `master` is a *good* practice, however it must be done *right*. Merging *master* into the feature branch creates conflicts for git.

One useful piece of *git advice*: you should *rebase* your branch on the master branch instead of *merging* the master branch into yours. (Or in a Continuous Delivery workflow, you would merge *your* branch *into master*...but this is not relevant here).

If you are used to `git pull`, you can rebase instead with the following command `git pull --rebase`.

U8NWXD:
I think PassFF wasn't doing anything because I deleted the `return result` lines in `getPasswordData` when removing my implementation in `pass.js`. I fixed this in be3857b. Thanks for your advice about rebasing! I'll look into re-doing that merge to resolve the conflicts.

5bentz:
Well, you see now the consequences of your merge, haha.

Another point: Commit messages should respect the following syntax

```
Short (50 chars or less) summary of changes

More detailed explanatory text, if necessary. Wrap it to
about 72 characters or so. In some contexts, the first
line is treated as the subject of an email and the rest of
the text as the body. The blank line separating the
summary from the body is critical (unless you omit the body
entirely); tools like rebase can get confused if you run
the two together.

Further paragraphs come after blank lines.

 - Bullet points are okay, too
 
 - Typically a hyphen or asterisk is used for the bullet,
preceded by a single space, with blank lines in
between, but conventions vary here
```
Reference: ProGit page 127
You can download the book here: https://git-scm.com/doc
This is your one-stop book for git :book: 

For the rebase, I would suggest:

For an introduction to git rebase: [this dev.to article](https://dev.to/maxwell_dev/the-git-rebase-introduction-i-wish-id-had) or ProGit (page 91)

1. Create a new branch
  A rebase can go *wrong* or might not be doing what you expect, so it is better and *free* to create a new branch and experiment on it.
  `git checkout -b rebase_br domain_check`
2. Rebase
  `git rebase --onto master 80b37bf`
  80b37bf is “Merge remote-tracking branch 'passff/master'”
3. Drop the commits than tried to fix the mess
```
pick ad45864 Write method to perform security checks on domain before autofilling passwords
pick 37fd678 Add domain security check whenever getting password data
pick 6769190 Make warnings consistently display URL/domain
pick 1f6a254 Add semicolons and restructure warning messages
pick 6e2ecd9 Make warning messages locale-adaptive
pick 4093253 Replace override question constant with locale-specific strings
pick 35dfd1c Remove empty line
pick 40014ae Add preferences option (off by default) for domain checks
pick 525b6df Split function domainSecurityCheck
pick b4d0673 Move security checks to page.js
pick 3224a56 Security checks: misc improvements
drop 8e75316 Remove security checks from pass.js
pick be04dfd Make security check code more readable
pick 83db464 Make anti-phishing check messages clearer
drop be3857b Fix Broken Password Retrieval When moving the domain security check functionality from `pass.js` to `page.js`, the `return result` lines in `getPasswordData` were accidentally deleted as well. This commit adds them back.
pick 697cbaf Clean up UI string formatting by passing parameters

# Rebase 80b37bf..697cbaf onto 02a448a (16 commands)
```
4. Save & quit
`:wq` in VIM
5. Success!? Not yet
`Successfully rebased and updated refs/heads/rebase_br`
6. Check that the rebase has gone right
If the rebased branch has greatly diverged from the old branch, something might be fishy.
`git diff domain_check`
We expect no diff, or minor diff such as whitespaces.
7. A more thorough check
  * The graph: `git log --oneline --decorate --graph --all `
  * The code: `git log -p`

I have done it and it went well: PassFF works :+1:

U8NWXD:
I think my branch has now been rebased correctly, but please let me know if problems remain.

U8NWXD:
@5bentz Did I rebase correctly?

5bentz:
@U8NWXD Oh sorry.
You did it good :+1: There is no merge conflict anymore.

tuxor1337:
So far, we use our own function `Page.notify(...)` to inform users about stuff. This is mainly because, this way, we can style the notification dialog in such a way that the user recognizes it as coming from PassFF. A generic `window.alert` or `window.confirm` can be irritating because it is styled like any dialog from the web page.

If you agree with this point of you we could think about replacing all of the calls to `window.alert` and `window.confirm` with our own implementations. So, what do you think?

U8NWXD:
I agree that the generic alerts and confirmation dialogs are unappealing. I used `window.alert` and `window.confirm` because they block following code from executing until the dialog has been resolved. The blocking functionality let me insert the checks into the existing code without rearrangement.

I don't think blocking is possible with custom dialogs, so I'll look into rearranging the existing code to work with custom dialogs. With the changes 5bentz suggested, I think this would be easier now than when I initially decided against it. Here is the structure I am considering:
```
function askUser(prompt, proceed, cancel) {
  Create dialog displaying prompt and having buttons to abort or continue.
  The continue button calls proceed(...), and the abort button calls cancel(...)
}
function proceed(...) {
  Fill Password
}
function cancel(...) {
  Abort Filling
}
if (securityChecks(...) are OK) {
  proceed(...)
} else {
  askUser("Some Warning", proceed, cancel)
}
```
I am not sure how to handle the `return passwordData;` at the end of `fillInputs: content_function("Page.fillInputs", function (item, andSubmit)` in `page.js`. What is being done with `passwordData` later, and can that be replicated in the structure I propose above?

tuxor1337:
This kind of problem is handled using promises in most parts of PassFF. In this case, the security checks should probably return promises that are resolved with the result of the (interactive) checks.

However, we have to rewrite `Page.notify` to return a promise that is resolved once the user presses OK and we have to implement a function `Page.confirm` that creates a dialog with OK and Cancel buttons.

PR #335, opened 2018-06-10
Prompt for username/url in generate dialog

eklitzke:
This adds a prompt for a username/url in the generate dialog. The input fields look like this: https://monad.io/mtdflgki.png

Given those inputs, the entry in the pass manager will end up looking like this:
```
vt&M~BWT#]E.u#2_,
---
test
password: vt&M~BWT#]E.u#2_,
username: test-user
url: test-url
```

If the username/url are omitted then the password manager acts just as it does now.

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -173,6 +173,12 @@
   "passff_newpassword_inputs_password_confirmation_label": {
     "message": "Confirm password"
   },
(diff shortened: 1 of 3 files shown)


tuxor1337:
Thanks for the PR!

Can you explain, why you put that `---` line followed by lines for name and password (which are both already in the file name and the file's first line).

tuxor1337:
This is obsolete due to PR #379

PR #339, opened 2018-06-13
Fix invisible input menu's icon

5bentz:
Input fields with the style '-moz-appearance: window' hide the PassFF icon.

Minimal Working Example

- copy-paste the HTML code below in a HTML file
- open it with Firefox
```
<html>
    <head>
        <meta charset="utf-8" />
    </head>
    <body>
        <input name="login" type="text" style="-moz-appearance:window;">
    </body>
</html>
```

I have also included `-webkit-appearance:none` for the other browsers (anticipating the chrome port...).

Tested on:
* Firefox 52 ESR and Firefox 61.0b13 Dev Edition.
* Linux Debian Stretch

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -241,6 +241,8 @@ PassFF.Page = (function () {
     if (typeof input.passff_injected !== "undefined") return;
     log.debug("Inject icon", input.id || input.name);
     input.passff_injected = true;
+    input.style["-moz-appearance"] = "none";
+    input.style["-webkit-appearance"] = "none";
     input.style.backgroundRepeat = "no-repeat";
     input.style.backgroundAttachment = "scroll";
     input.style.backgroundSize = "16px 16px";


tuxor1337:
Are there actual websites that are affected by this?

5bentz:
Yes [there is](https://animedigitalnetwork.fr/index.php/connexion)
I'd like to find statistics about its use and I failed to do so :x

tuxor1337:
Interesting - we learn from that example that `-moz-appearance: window;` actually affects the layout. If you remove it and click into the field, the background color changes while it stays the same otherwise. Therefore, I don't think it's safe to just enforce `-moz-appearance: none;` since it might break the layout completely in some examples.

Do you know what exactly happens or which aspect of the layout are affected when setting `-moz-appearance: window;`?

5bentz:
MDN has a page about `-moz-appearance`, but nothing about the `window` value. This property shouldn't be used (it is experimental & non-standard). Interesting...
https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance

While experimenting with the code snippet above in my browser (FF 52 & 61), I see that `-moz-appearance: window;` affects:
* its dimension:  `border`s are set to 0 (from 2 on FF52, 8 on FF61)
* its skin: it becomes a solid grey block. The "appearance" depends on the configuration of the OS, thus it *should* vary across OS or Desktop environments...From MDN: *display an element using platform-native styling based on the operating system's theme.*

Using this property must be clumsiness.

WTF?! I was writing this message and I somehow merged the PR...Okay...We can still revert it...
GUI FTW.

tuxor1337:
What I want to say is that setting `-moz-appearance: window` affects the background color. If that is intended by the webdesigner, we will break his layout for the user by setting `-moz-appearance: none`. Furthermore, we would force `-moz-appearance: none` even for cases where `-moz-appearance` has a value different from `window`.

I illustrate this in the following example: https://jsfiddle.net/x3wgmn6z/16/ There, all input fields of class `special` have the property `-moz-appearance: window` and I simulate the override of passff by enforcing `-moz-appearance: none` for the third field. As a side effect, the user is not able to read the contents of this input field any more.

One possible way to deal with this would be to do something similar to the fourth input field in the example: We could *check* whether the property `-moz-appearance: window` is set and **only if exactly that's the case** we would set `-moz-appearance: none` and *in addition* simulate the behavior by setting `background-color: #e9e9e9; border: 0;`.

However, this seems awkward because we try to make up for bad (or at least weird) webdesign - we don't even know for how many pages out there this applies. Moreover, the effect of `-moz-appearance: window` might change in the future since it's undocumented and non-standard.

My suggestion: Write a mail or message to ADN notifying them about the problem and suggesting to use web standards.

tuxor1337:
Note that the way we currently inject the PassFF icon is already problematic: If the web designer has already set a background-image we *override* his settings. Instead, we should rather have the PassFF icon display *in addition* - e. g., by using multiple CSS backgrounds: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Backgrounds_and_Borders/Using_multiple_backgrounds

PR #342, opened 2018-06-21
Check host app version by semantic versioning

5bentz:
See https://github.com/passff/passff-host/issues/18

To be able to bugfix the host app without touching PassFF, we have to enforce *semantic versioning* in passff-host.
As a result, PassFF should more loosely check passff-host's version.

Note: The lexicography check requires its operands to be `string`s. Thus, I explicitly coerce `version` to be a `string`.

#### What is left to do?
Let the user know that new versions of passff-host are available.
This will be the object of another PR.

### DIFF
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -230,12 +230,12 @@ PassFF.Pass = (function () {
             let version = result.version || "0.0";
             const compatible = (function isHostAppCompatible(version) {
               const MIN_VERSION = '1.0.1';
-              return semver().gte(version, MIN_VERSION) || version === "testing";
+              return version === "testing" || semver.gte(version, MIN_VERSION);
             })(version);
             if (!compatible) {
               log.warn("The host app is outdated!", version);
               result.exitCode = -2;
-              result.stderr = "The host app is outdated! v" + version;
+              result.stderr = `The host app (v${version}) is outdated!`;
             } else if (result.exitCode !== 0) {
               log.warn('Script execution failed',
                 result.exitCode, result.stderr, result.stdout);
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -227,9 +227,12 @@ PassFF.Pass = (function () {
         }
         return browser.runtime.sendNativeMessage("passff", args)
           .then((result) => {
-            let version = String(result.version) || "0.0";
-            const COMPATIBLE_VERSION = "1.0.1";
-            if (version < COMPATIBLE_VERSION && version !== "testing") {
+            let version = result.version || "0.0";
+            const compatible = (function isHostAppCompatible(version) {
+              const MIN_VERSION = '1.0.1';
+              return semver().gte(version, MIN_VERSION) || version === "testing";
+            })(version);
+            if (!compatible) {
               log.warn("The host app is outdated!", version);
               result.exitCode = -2;
               result.stderr = "The host app is outdated! v" + version;
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -227,11 +227,12 @@ PassFF.Pass = (function () {
         }
         return browser.runtime.sendNativeMessage("passff", args)
           .then((result) => {
-            let version = result.version || "0.0";
-            if (version !== "1.0.1" && version !== "testing") {
+            let version = String(result.version) || "0.0";
+            const COMPATIBLE_VERSION = "1.0.1";
+            if (version < COMPATIBLE_VERSION && version !== "testing") {
               log.warn("The host app is outdated!", version);
               result.exitCode = -2;
-              result.stderr = "v" + version;
+              result.stderr = "The host app is outdated! v" + version;
             } else if (result.exitCode !== 0) {
               log.warn('Script execution failed',
                 result.exitCode, result.stderr, result.stdout);
--- a/src/modules/util.js
+++ b/src/modules/util.js
@@ -175,7 +175,7 @@ function fun_name(name) {
  * See https://github.com/passff/passff/pull/342
  */
 
-function semver() {
+const semver = (function semver() {
 
   /**
    * Takes a version string as input and parses it.
@@ -183,12 +183,13 @@ function semver() {
    */
   function parser(version) {
     version = version + "";
-    // This regEx is ^(id)\.(id)(?:\.(id))?(?:ext)$
-    // with id = `0|[1-9]\d*` and ext = `|[\+-].+`
-    const reg = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].+)$/;
-    const m = version.match(reg);
 
-    if (!m) return null;
+    const id = String.raw`(0|[1-9]\d*)`;
+    const ext = String.raw`(?:|[\+-].+)`;
+    const reg = String.raw`^${id}\.${id}(?:\.${id})?${ext}$`;
+    const m = version.trim().match(reg);
+
+    if (!m) throw new TypeError('Invalid Version: ' + version);
 
     return {
       major: +m[1] || 0,
@@ -210,8 +211,6 @@ function semver() {
     v1 = parser(v1);
     v2 = parser(v2);
 
-    if(!v1 || !v2) return NaN;
-
     return (v1.major - v2.major ||
             v1.minor - v2.minor ||
             v1.patch - v2.patch);
@@ -229,8 +228,6 @@ function semver() {
     return comparator(v1,v2) == 0;
   }
 
-  // The API is conservative for now.
-  // It could be expanded later if necessary.
   let publicAPI = { gt, gte, eq };
   return publicAPI;
-}
+})();
--- a/src/modules/util.js
+++ b/src/modules/util.js
@@ -161,3 +161,76 @@ function fun_name(name) {
     return [PassFF, parts[0]];
   }
 }
+
+/* #############################################################################
+ * #############################################################################
+ *  semantic versioning
+ * #############################################################################
+ */
+
+/*
+ * This implementation has been inspired by npm's semver.
+ *
+ * For in-depth explanations,
+ * See https://github.com/passff/passff/pull/342
+ */
+
+function semver() {
+
+  /**
+   * Takes a version string as input and parses it.
+   * Returns the major, minor and patch identifiers of the version as integers.
+   */
+  function parser(version) {
+    version = version + "";
+    // This regEx is ^(id)\.(id)(?:\.(id))?(?:ext)$
+    // with id = `0|[1-9]\d*` and ext = `|[\+-].+`
+    const reg = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].+)$/;
+    const m = version.match(reg);
+
+    if (!m) return null;
+
+    return {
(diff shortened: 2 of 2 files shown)


src/modules/pass.js
```diff
@@ -227,11 +227,12 @@ PassFF.Pass = (function () {
         }
         return browser.runtime.sendNativeMessage("passff", args)
           .then((result) => {
-            let version = result.version || "0.0";
-            if (version !== "1.0.1" && version !== "testing") {
+            let version = String(result.version) || "0.0";
```
tuxor1337: This will never fall back to "0.0", but you might end up with version strings like "undefined" or "null". How about `String(result.version || "0.0")`?

src/modules/pass.js
```diff
@@ -227,11 +227,12 @@ PassFF.Pass = (function () {
...
+            const COMPATIBLE_VERSION = "1.0.1";
+            if (version < COMPATIBLE_VERSION && version !== "testing") {
```
tuxor1337: I don't think that this check does what you want. Just look at `"1.9" < "1.10"`which returns `false`

5bentz:
After reviewing your URLs, and other implementations & licenses:
* https://mail.mozilla.org/pipermail/dev-addons/2017-June/002916.html # there is no built-in comparator in Firefox anymore...
  * https://gist.github.com/cowboy/566233/a2025cf4fa24668a57d13926ec181de4540c6e4d
    * we have to avoid `eval()` and `new Function(..)` because of Firefox's CSP (CSP [*could*](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_Security_Policy) be disabled though)
* https://opensource.guide/legal/ # we can use code licensed under GPL2 or under a permissive license


I'd say we have two solutions:

# Use a well known & tested code

For this purpose, npm's [semver](https://github.com/npm/node-semver) licensed under ISC (permissive) looks perfect:
  * '+ compatible
  * '+ has all the functions we need (and way more)

We can either import it manually (copy-paste) or use npm.

Other candidates:

* https://gist.github.com/cowboy/566233/a2025cf4fa24668a57d13926ec181de4540c6e4
  * '+ compatible (under GPLv2)
  * '+ validate the version strings
  * '+ nicer syntax (v1 <= v2)
  * '- use of `new Function()`

* https://stackoverflow.com/a/16187766
  * '+ very simple
  * '- not compatible (under CC-BY-SA 3.0)
  * '- lack essential functions (parser, greater than)


# Code it ourselves

While npm's semver looks perfect, it is way too elaborated for us, which is not really an issue.

On the other hand, having our own implementation could help understand how npm's semver works. And ultimately, we could always switch to npm's semver.

Actually, a semver parser for our needs (x.y.z) is quite simple, but not obvious.

I recommend you to look at npm's semver implementation which inspired mine. The code is one thousand lines long, however each regexp or function is really simple and explicit. I will link some lines throughout the explanation below.

/!\ The following looks more like an average dev.to article about semver than anything else! Which is nice.

We need:
1. a parser of version strings `x.y.z`
2. a comparator of semantic versions

## The parser

The parser's goal is twofold:
  1. validate what is a good version string
  2. extract the identifiers MAJOR MINOR PATCH.

### 1. What is a good version string?

As far as we're concerned, these are 1.0, 1.0.1, 1.0.2, 1.1, 1.1.42, 1.28.0, etc...
Formally, a good version is: one 'identifier', followed by one 'dot identifier', and maybe followed by a second 'dot identifier'. 

With a RegEx `(id)\.(id)(?:\.(id))?` (I'll tackle `?:` later).
  * `(id)` is the first identifier
  * `\.(id)` the second
  * `(?:\.(id))?` the optional third

To accept pre-release & build-metadata, we coukd add the RegExp `(?:|[\+-].*)` which means: *Nothing, or anything preceded by a sign `+` or `-`*. This is loose, but enough.

Finally, *anchor* it with `^.. $`: `^(id)\.(id)(?:\.(id))?(?:|[\+-].*)$`

#### Hey! What is an identifier `id`? 

I'm glad you asked! :)

Semver's npm definition is pretty straightforward.
> A single `0`, or a non-zero digit followed by zero or more digits.

See: https://github.com/npm/node-semver/blob/master/semver.js#L36

In RegExp: `0|[1-9]\d*`

Thus, after literally replacing `id` with `0|[1-9]\d*`, our RegExp becomes:
```
^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].*)$
```

### 2. Extract the identifiers

It is already done thanks to *parenthesized substring matches*.
Remember the `?:`? All matches in parentheses can be reused later, except matches in `(?:)`, called *non-capturing parentheses*.

Our RegExp uses capturing parentheses `( )` ONLY for the identifiers.

Thus, we can retrieve them back using: [`String.prototype.match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match).

See: https://github.com/npm/node-semver/blob/master/semver.js#L302

/!\ These are quite complete (and JS-compliant) pseudo-code.

```
parser(version) {
  reg = ..;
  m = version.match(reg);
  
  if (!m)
    return null;

  return {
    major: +m[1] || 0
    minor: +m[2] || 0
    patch: +m[3] || 0
  };
}
```  

## The comparator

Inspired by: https://stackoverflow.com/a/16187766

Such a comparator returns a number: (notice the symmetry in the relations)
* == 0 if v1 == v2
* < 0 if v1 < v2
* '> 0 if v1 > v2

For *free*, we obtain a function to sort versions in the following manner: `array.sort(comparator)`

```
comparator(v1,v2) {
  v1 = parser(v1)
  v2 = parser(v2)
  
  if(!v1 || !v2) // if one string is invalid, returns `null` which is coerced to `false`.
    return null
  
  return (v1.major - v2.major
          || v1.minor - v2.minor
          || v1.patch - v2.patch);
}

Then, we can easily define more useful operators to compare versions:

ge(v1, v2) { // Greater or Equal
  return comparator(v1,v2) >= 0;
}
```

## The actual comparison in the code

Current code: https://github.com/passff/passff/blob/master/src/modules/pass.js#L231

If the version is compatible or "testing", then the host app is accepted.
There is no change in the logic of the current code on the master branch.

```
isHostAppCompatible(version) {
  COMPATIBLE_VERSION = '1.0.1';
  return ge(version, COMPATIBLE_VERSION) || version === "testing";
}


if(!isHostAppCompatible(version))
  warn_app_outdated;
else
  do_stuff;
```


# What needs to be done?

I hope it was interesting. What do you think?

tuxor1337:
Very well written! I would be fine with an implementation along the lines of what you describe because I don't like the legal and packaging overhead that comes with including third-party code into the project. As you say, we don't really need the flexibility of `semver.js`.

Under `src/modules/util.js` we already have a general purpose collection of utility functions. A version parser and comparator would fit into that collection quite well I think. If you wish, please finalize your implementation as part of this PR.

5bentz:
The PR is here!
It has been implemented in `util.js` as a module *old-style* (pre-ES6).

Differences from the version above:
* `parser` explicitly coerces to `string`.
* the regExp only accepts extensions with *non-empty* identifiers
* `comparator` returns `NaN` instead of `null`. `null` coerces to 0 which is equality...not good. [1]
* `ge` has been renamed `gte` to follow npm's terminology. 

[1] Should we throw a `TypeError` instead?

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+ */
+
+function semver() {
+
+  /**
+   * Takes a version string as input and parses it.
+   * Returns the major, minor and patch identifiers of the version as integers.
+   */
+  function parser(version) {
+    version = version + "";
```
tuxor1337: As you say, why not have a check
```
if (typeof version !== 'string') {
  throw new TypeError('Invalid Version: ' + version);
}
```

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+    // This regEx is ^(id)\.(id)(?:\.(id))?(?:ext)$
+    // with id = `0|[1-9]\d*` and ext = `|[\+-].+`
+    const reg = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].+)$/;
+    const m = version.match(reg);
+
+    if (!m) return null;
```
tuxor1337: Again, IMO a type error would be nice.

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+  /**
+   * Takes a version string as input and parses it.
+   * Returns the major, minor and patch identifiers of the version as integers.
+   */
+  function parser(version) {
+    version = version + "";
+    // This regEx is ^(id)\.(id)(?:\.(id))?(?:ext)$
+    // with id = `0|[1-9]\d*` and ext = `|[\+-].+`
+    const reg = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].+)$/;
+    const m = version.match(reg);
```
tuxor1337: I suggest using `version.trim().match(reg)` as in the npm semver.

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+   *   - positive (>0) if v1 > v2
+   *   - zero     (=0) if v1 = v2
+   * Notice the symmetry in the relations.
+   * Last but not least, we can sort versions using Array#sort(comparator).
+   */
+  function comparator(v1,v2) {
+    v1 = parser(v1);
+    v2 = parser(v2);
+
+    if(!v1 || !v2) return NaN;
```
tuxor1337: See above, I think the parser should sort this out so that we can cancel this check altogether.

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+  function gte(v1, v2) {
+    return comparator(v1,v2) >= 0;
+  }
+
+  function eq(v1, v2) {
+    return comparator(v1,v2) == 0;
+  }
+
+  // The API is conservative for now.
+  // It could be expanded later if necessary.
```
tuxor1337: I don't think the comment is necessary.

src/modules/util.js
```diff
@@ -161,3 +161,76 @@ function fun_name(name) {
...
+
+  /**
+   * Takes a version string as input and parses it.
+   * Returns the major, minor and patch identifiers of the version as integers.
+   */
+  function parser(version) {
+    version = version + "";
+    // This regEx is ^(id)\.(id)(?:\.(id))?(?:ext)$
+    // with id = `0|[1-9]\d*` and ext = `|[\+-].+`
+    const reg = /^(0|[1-9]\d*)\.(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:|[\+-].+)$/;
```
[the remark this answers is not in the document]
5bentz: I'll go for
```
reg = String.raw`^${id}\.${id}(?:\.${id})?${ext}$`;
```
`String.raw` frees us from the necessity of escaping RegExp escapes: such as `\d \.`

:+1: 
C'mon I'm at chapter 1 of [*You Don't Know JS: ES6 & Beyond*](https://github.com/getify/You-Dont-Know-JS). String literals are presented in chapter 2!

5bentz:
Thanks for the review, changes pushed!

PR #359, opened 2018-08-23
Move Makefile to root

shtrom:
Signed-off-by: Olivier Mehani <shtrom@ssji.net>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 
 VERSION := testing
 
-SRC_DIR := .
-TARGET_DIR := ../bin/$(VERSION)
-BUILD_DIR := ../build/$(VERSION)
+SRC_DIR := ./src
+TARGET_DIR := ./bin/$(VERSION)
+BUILD_DIR := ./build/$(VERSION)
 
 CONTENT_FILES := content _locales modules skin icon.png
 CONTENT_SRC := $(addprefix $(SRC_DIR)/,$(CONTENT_FILES))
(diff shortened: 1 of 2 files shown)


tuxor1337:
Does this go back to some kind of standard or best-practice? If so, can you give a reference?

shtrom:
I don't have a proper reference, but I can give you a bandwagon fallacy to support it.
* https://github.com/torvalds/linux/blob/master/Makefile
* https://github.com/gcc-mirror/gcc/blob/master/Makefile.in (Makefile gets generated by the configure script in the same directory)
* https://github.com/GNOME/gimp/blob/master/Makefile.am (ditto by automake+configure)
* https://github.com/LibreOffice/core/blob/master/Makefile.in
* https://github.com/openbsd/src/blob/master/Makefile
* https://github.com/openssh/openssh-portable/blob/master/Makefile.in
* https://hg.mozilla.org/mozilla-central/file/tip/Makefile.in
* https://github.com/nodejs/node/blob/master/Makefile

Heh, it's also an appeal to authority. Two fallacies in one argument and it's not even 9AM!

FWIW, I think this shows it is common practice: in my search of big FOSS projects from the top of my head, the only one that I did not include in the list above was GTK, simply because it uses a different build/bootstrap system (for which configuration files are also at the root).

tuxor1337:
Hm, that's not really satisfactory, especially because most of these use the GNU coding standard that suggests the use of a `configure` script and on the fly Makefile generation. In fact, the generated makefiles might end up in a different place depending on the directory variables.

However, if you prefer the Makefile in the root directory, I'm also fine with it ;)

PR #363, opened 2018-08-30
We can't really have the PSL (yet?) but let's offer users a stop gap:…

tialaramex:
Please be gentle, literally the first time I've written Javascript in anger unless DOM mangling in like Mozilla M14 or whatever that was back at the turn of the century counts.


tuxor1337:
Thanks a lot!

PR #364, opened 2018-09-02
fix z-index issue on e.g. amazon.de login

girst:
I've noticed a bug with the (P) popup. Currently, only the z-index of the parent elements are checked, but not of elements nearby (which the popup may overlap). On amazon.de's login page for example they set the 'Anmelden' (login) button to z-index:20, but our popup only has the default z-index of 1. 

This patch uses querySelectorAll to filter through all z-indices and gives the popup the highest found. 

Is there a reason the addon only checked the parents? performance, maybe?

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -349,8 +349,8 @@ PassFF.Page = (function () {
     popup_menu.style.display  = "block";
 
     // get the largest z-index value and position ourselves above it
-    let z = [...document.querySelectorAll('*')]
-      .filter(e => e.style.position !== "static")
+    let z = [...document.querySelectorAll('body *')]
+      .filter(e => !(e.style.position in ["static", ""]))
       .map(e => window.getComputedStyle(e).zIndex)
       .filter(e => e>0)
       .sort()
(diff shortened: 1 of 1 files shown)


tuxor1337:
Yes, calling "getComputedStyle" on all elements is a potential performance bottle neck. Can you at least filter by `el.style.position !== "static"`, please?

girst:
i see. done!

girst:
can you have a look again, please?

> This is not how you use the JavaScript "in" operator (maybe you confuse it with Python?). We want to have something like `["static",""].indexOf(...) === -1`.
oops! yeah, my js is all but great and I've done too much python at $WORK :/

> We should call `parseInt` since `zIndex` returns a string and for strings, `sort()` does lexical sorting (`"100" < "22"`). When `parseInt` returns `NaN`, we should default to `1`.
d'oh! nans are removed with the e>0 filter below.

> We will get better performance when using `Math.max` instead of `Array.sort`. Make sure to include 1 into `Math.max` so that we don't get `-Infinity` for empty Arrays.
also done. (how didn't i think of this?!)

PR #379, opened 2019-01-14
Prompt for additional info on generate

BenjaminDebeerst:
The additional info is added to the new pass entry right after creating
it in a separate step because `pass generate` has no option to append
data to the generated password during creation.

This PR is heavily based on the work by @eklitzke in #335. This PR is a (from my point of view) improved version that I hope will be merged rather than linger around without activity. I am new to JS and Firefox plugin development. So any feedback, both regarding functionality and style, is greatly appreciated.

I felt that allowing any additional data has many advantages over two very specific fields:

* Users can choose and use their own naming scheme
* Users can add more than only two rows
* This greatly reduces the code complexity because we now can reuse the `addNewPassword` directly.

### DIFF
--- a/src/content/passwordGenerator.html
+++ b/src/content/passwordGenerator.html
@@ -25,6 +25,9 @@
             <input type="checkbox" id="gen-include-symbols" checked="true" />
             <label for="gen-include-symbols">inputs_include_symbols_label</label>
           </p><p>
+            <label>inputs_additional_info_label</label>
+            <textarea id="add-additional-info-generate"></textarea>
+          </p><p>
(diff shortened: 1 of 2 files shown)


tuxor1337:
Yes, I certainly prefer this over #335. The implementation looks okay to me. Unfortunately, I don't have time to test it. Maybe someone else can confirm that this works as it's supposed to?

ccb23:
I just did a very basic test. I patched passff@invicem.pro.xpi in my firefox profile directory with the diffs from this PR. And it seems to work for me :-)

Issue #33, opened 2014-09-06
[AMO] subprocess


jvenant:
Since `subprocess.jsm` does not restore the process working directory after changing it to accommodate the `workdir` parameter, you must manually restore it when using it as such.

PR #384, opened 2019-02-05
OTP support

### LINKED ISSUE #200: Add support for pass-otp plugin
It would be amazing if we could add support for [pass-otp plugin](https://github.com/tadfisher/pass-otp) also, which provides OTP ability.

LucidOne:
This provides support for [pass-otp](https://github.com/tadfisher/pass-otp) and resolves #200 however it requires [passff-host #33](https://github.com/passff/passff-host/pull/33) to be updated as well.

:city_sunrise:

### DIFF
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -254,9 +254,13 @@ PassFF.Pass = (function () {
               result.exitCode = -2;
               result.stderr = `The host app (v${version}) is outdated!`;
             } else if (result.exitCode !== 0) {
-              log.warn('Script execution failed',
-                result.exitCode, result.stderr, result.stdout);
-              PassFF.Page.notify("Script execution failed: \n" + result.stderr);
+              if (result.stderr.trim() == "Error: /otp is not in the password store.") {
+                PassFF.Page.notify("OTP Support requires Host app version 1.1.0 or later.\nhttps://github.com/passff/passff-host/");
+              } else {
+                log.warn('Script execution failed',
+                  result.exitCode, result.stderr, result.stdout);
+                PassFF.Page.notify("Script execution failed: \n" + result.stderr);
+              }
             } else {
               log.debug('Script execution ok');
             }
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -48,6 +48,15 @@ PassFF.Pass = (function () {
     passwordData.url = url;
   }
 
+  function setOtpauth(passwordData) {
+    let otpauth;
+    for (let i = 0; i < PassFF.Preferences.otpauthFieldNames.length; i++) {
+      otpauth = passwordData[PassFF.Preferences.otpauthFieldNames[i]];
+      if (otpauth) break;
+    }
+    passwordData.otpauth = !!otpauth;
+  }
+
   function setOther(passwordData) {
     let other = {};
     Object.keys(passwordData)
@@ -455,14 +464,18 @@ PassFF.Pass = (function () {
             setLogin(result, item);
             setPassword(result);
             setUrl(result);
+            setOtpauth(result);
             setOther(result);
             setText(result, executionResult.stdout);
 
-            return this.generateOtp(key)
-              .then((otp) => {
-                result.otp = otp;
-                return result;
-              });
+            if (result.otpauth) {
+              return this.generateOtp(key)
+                .then((otp) => {
+                  log.debug('Generating OTP token');
+                  result.otp = otp;
+                  return result;
+                });
+            }
 
             return result;
           });
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -236,7 +236,7 @@ PassFF.Pass = (function () {
       function (args) {
         let command = "ls";
         if (args.length > 0) {
-            if (["insert","generate"].indexOf(args[0]) >= 0) {
+            if (["insert","generate","otp"].indexOf(args[0]) >= 0) {
                 command = args[0];
             } else {
                 command = "show";
@@ -253,14 +253,13 @@ PassFF.Pass = (function () {
               log.warn("The host app is outdated!", version);
               result.exitCode = -2;
               result.stderr = `The host app (v${version}) is outdated!`;
+            } else if (command == "otp" && semver.gt("1.1.0", version)) {
+              log.warn("This version of the host app does not support OTP!", version);
+              PassFF.Page.notify("OTP Support requires Host app version 1.1.0 or later.\nhttps://github.com/passff/passff-host/");
             } else if (result.exitCode !== 0) {
-              if (result.stderr.trim() == "Error: /otp is not in the password store.") {
-                PassFF.Page.notify("OTP Support requires Host app version 1.1.0 or later.\nhttps://github.com/passff/passff-host/");
-              } else {
-                log.warn('Script execution failed',
-                  result.exitCode, result.stderr, result.stdout);
-                PassFF.Page.notify("Script execution failed: \n" + result.stderr);
-              }
+              log.warn('Script execution failed',
+                result.exitCode, result.stderr, result.stdout);
+              PassFF.Page.notify("Script execution failed: \n" + result.stderr);
             } else {
               log.debug('Script execution ok');
             }
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -457,6 +457,13 @@ PassFF.Pass = (function () {
             setUrl(result);
             setOther(result);
             setText(result, executionResult.stdout);
+
+            return this.generateOtp(key)
+              .then((otp) => {
+                result.otp = otp;
+                return result;
+              });
+
             return result;
           });
       }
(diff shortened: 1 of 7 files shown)


LucidOne:
I just got it "working" so I thought I'd post up what I have. After some :sleeping: I'm happy to clean up the pull request.

What I think it needs to do is check the return from `pass show` for the `otpauth://` field before generating the OTP token.

LucidOne:
Updated the pull request to avoid the second call to the host app and added OTP information to the README.

tuxor1337:
Ok, thanks! No we need a third person to test this :)

madduck:
I tried this, but upon attempting to fill in a form, I get

```
Script execution failed:
Error: /otp is not in the password store.
```

I don't really have time to look into this right now. But happy to try again.

LucidOne:
It sounds like [passff-host](https://github.com/passff/passff-host) needs to be updated.

madduck:
Argh, I had done that… but on the wrong computer! ;)

Works now. Very nice. Anything in particular that I should test?

LucidOne:
I only tested on github  and sr.ht, so if you can test any other sites that would be great.
One site named the Input `code` and the named it `otp`, I'm not sure what else is common.

madduck:
I tested on bitcoin.de and poloniex.com, as well as Twitter, and two internal sites.

tuxor1337:
Thanks for testing!

Now I'm wondering what's the best way to ship this change. The host app is 100% backward compatible (apart from the rare case that someone happens to have an entry called "otp" in the root of her password store...). But once we add the OTP feature to the extension, we either have to enforce a more recent version of the host app (https://github.com/passff/passff/blob/master/src/modules/pass.js#L240) or we automatically deactivate the OTP feature whenever we detect that an older version of the host app is in use. The first option is a pain for most users (namely those who don't care about OTP support).

Another option is to implement a self-update functionality into the host app: https://github.com/passff/passff-host/issues/34 This would require a manual update and after that, users will never have to deal with this problem again.

madduck:
I don't think self-update is a good idea. I think we should go for graceful fallback.

tuxor1337:
If we go for the gracefull fallback we should inform the user about this somewhere, e.g., in the preferences and in the README. Moreover, we should deactivate certain OTP related functionality like the "Add otp input name" in the context menu and we should make sure that `generateOtp(key)` isn't called.

LucidOne:
So if the extension is upgraded but the host app isn't, the current error message is
```
Script execution failed:
Error: /otp is not in the password store.
```
Looking back at my [implementation notes](https://github.com/passff/passff-host/pull/33#issuecomment-460535932).

I just tried running one of my tests on the old version of passff.py
```
$ echo -e "\x19\x00\x00\x00[\"otp\",\"/www/github.com\"]" | ./passff.py | tail -c +4; echo
{"exitCode": 1, "stdout": "", "version": "_VERSIONHOLDER_", "stderr": "Error: /otp is not in the password store.\n"}
```
So it _could_ be programmatically handled if exitCode == 1 and stderr == "Error: /otp is not in the password store.\n".

It only ever gets to this branch of code if you have otpauth in your passfile. I almost wonder if a javascript alert box is the right way to deal with this.

LucidOne:
Also, as for updating to host app. I'm looking at fixing [this](https://github.com/passff/passff-host/issues/1) the next time I have to install passff on another machine.

tuxor1337:
Do I understand correctly that only users that are probably interested in using the new OTP feature will be adversely affected by this? And those users will be okay with making an effort to update the host app this time, because it brings a benefit for them. Therefore we don't enforce the most recent version of the host app for all users, but we wait for an otpauth entry and then we catch the error message replacing it by: "Did you know that PassFF does support otp...? You just have to update your host app..."

LucidOne:
I think this is the case as I understand the initial error @madduck encountered was that he made it past the username/password page before having an issue trying to generate an otp token from the old version of passff-host.

The relevant code
```
  function setOtpauth(passwordData) {
    let otpauth;
    for (let i = 0; i < PassFF.Preferences.otpauthFieldNames.length; i++) {
      otpauth = passwordData[PassFF.Preferences.otpauthFieldNames[i]];
      if (otpauth) break;
    }
    passwordData.otpauth = !!otpauth;
  }
```
[This](https://github.com/passff/passff/pull/384/files#diff-58e1ee42318fc2d0f414913f5fe09e5dR51) sets `passwordData.otpauth` to `true` if there is a field in the passfile that matches "otpauth", as in `otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example`.

If `passwordData.otpauth` is True, [this](https://github.com/passff/passff/pull/384/files#diff-58e1ee42318fc2d0f414913f5fe09e5dR471) then generates the OTP Token and stuffs it in `passwordData.otp`, 

[This](https://github.com/passff/passff/pull/384/files#diff-e8d31f21df475f877ac2881d2336da60R190) could check if `passwordData.otp` is set, but if I recall correctly, the worst case is that it fills an OTP input field with `Undefined` when you tell it to fill the form.

tuxor1337:
> [This](https://github.com/passff/passff/pull/384/files#diff-e8d31f21df475f877ac2881d2336da60R190) could check if `passwordData.otp` is set, but if I recall correctly, the worst case is that it fills an OTP input field with `Undefined` when you tell it to fill the form.

But that would mean that user input for those fields will be silently overwritten. Doesn't sound very elegant...

LucidOne:
It currently only gets overwritten if you click the (p) icon for the field or if you have autofill all forms enabled. Though to clarify, I think it _should_ check if `passwordData.otp`` is set before trying to fill the form. I can add this and a `window.alert()` to upgrade the host app to the pull request tomorrow.

I went through all of the comments on #200 however it looked like most of the concerns were not an issue simply because every website I have encountered uses OTP as a second factor, so you have to also have a login/password. I'm happy to have additional testing to find if other edge cases exist.

Can you update the [release](https://github.com/passff/passff-host/releases) for the host app to 1.1.0 so I have a minimum version number to reference in the alert dialog? Thanks!

tuxor1337:
There you go: https://github.com/passff/passff-host/releases/tag/1.1.0

LucidOne:
Updated. :tada: One minor issue, it looks like PassFF.Page.notify() strips out html so there is no way to link to where to upgrade. :man_shrugging:

src/modules/pass.js
```diff
@@ -245,9 +254,13 @@ PassFF.Pass = (function () {
               result.exitCode = -2;
               result.stderr = `The host app (v${version}) is outdated!`;
             } else if (result.exitCode !== 0) {
-              log.warn('Script execution failed',
-                result.exitCode, result.stderr, result.stdout);
-              PassFF.Page.notify("Script execution failed: \n" + result.stderr);
+              if (result.stderr.trim() == "Error: /otp is not in the password store.") {
```
lanodan: Wouldn’t it be better to do a check against the version of passff-host, like few lines earlier? Specially as it relies on a side-effect.

LucidOne: I think the idea here is that the host app only really needs to be upgraded if you are using OTP functionality, once the required version gets bumped this could be removed. I'm also ok with just bumping the required version of the host app.

tuxor1337: I think @lanodan suggests replacing the `if (result.stderr.trim() == "...")` by `if (command == "otp" && semver.gt("1.1.0", version))`. This check should already be placed before the line 256 as additional `} else if ...`. By the way, please add "otp" to the list of supported commands in line 239.

lanodan:
> I think @lanodan suggests replacing the `if (result.stderr.trim() == "...")` by `if (command == "otp" && semver.gt("1.1.0", version))`.

Yup I meant this one

src/modules/pass.js
LucidOne: Thanks for the feedback!

LucidOne:
I think I have gotten all of the outstanding issues now.
Anyone else have a moment to test the current version?

madduck:
Works beautifully. Thanks for your diligence. I vote for the merge!

lanodan:
It looks like it doesn’t works in the following setup:
- `…/otp` (for example `/Forges/github.com/otp`)
- otpauth line in it: `otpauth://totp/totp-secret?secret=…\n`

Either by filling with `Forges/github.com/otp` or with `Forges/github.com`.

I use this setup so otpauth doesn’t get erased when I re-generate the password with pass(1). It seems to work fine on passff-host side, does it needs the file to have multiple lines, even ~useless ones?

madduck:
Why would `otpauth` get erased when you regenerate the password? To be honest, I think the file-based scheme you are using isn't really supported (I may be wrong), and maybe the better solution than to hope for a fix, or hack something up, is for you to change the way you update passwords? Have you checked out the `update` extension?

LucidOne:
https://github.com/passff/passff/pull/384/files#diff-04c6e90faac2675aa89e2176d2eec7d8R99

It is unsupported by this patch. I think it just needs something done with `getPasswordData` under `// hierarchical-style item`.

Issue #389, opened 2019-02-19
Fails to fill form for bitcoin.de, possibly OTP related


madduck:
I just encountered a problem similar to #387 on bitcoin.de, using PassFF 1.6.3, as well as passff-host at commit a37d504b9035d7fc7c04fea1d05835e18a3cee6d, i.e. master after 1.1.0 was released:

```
fishbowl:…a/native-messaging-hosts% grep otp passff.py       #(pts/6) 20 11:44:48.127
    elif receivedMessage[0] == "otp" and len(receivedMessage) == 2:
        opt_args = ["otp"]
```

The login form at bitcoin.de is not being filled in while the `otpauth://` line is present in the data file. It works when I remove the line.

Here is the log output:

```
11:36:03.814 [PassFF.background] 2 matches for https://www.bitcoin.de/en/login util.js:66:9
11:36:04.887 [PassFF.background] Awaiting tab init... 88 https://www.bitcoin.de/en/login util.js:66:9
11:36:04.903 [PassFF.background] Tab init done 88 https://www.bitcoin.de/en/login util.js:66:9
11:36:05.717 [PassFF.background] Awaiting tab init... 88 https://www.bitcoin.de/en/login util.js:66:9
11:36:05.833 Error: WebExtension context not found!
ExtensionParent.jsm:1039:13
11:36:06.272 [PassFF.background] Message Preferences.getBrowserCommand received in background util.js:66:9
11:36:06.553 [PassFF.background] Message Pass.loadItems received in background util.js:66:9
11:36:06.659 [PassFF.background] Message Preferences.getKeyboardShortcut received in background util.js:66:9
11:36:06.675 [PassFF.background] Tab init done 88 https://www.bitcoin.de/en/login util.js:66:9
11:36:07.504 Error: WebExtension context not found!
ExtensionParent.jsm:1039:13
11:36:09.070 [PassFF.content] 2 matches for https://www.bitcoin.de/en/login util.js:66:9
11:36:09.073 [PassFF.content] Best fit item bitcoin.de/XXXXXX for https://www.bitcoin.de/en/login util.js:66:9
11:36:09.078 [PassFF.content] Inject icon login_username util.js:66:9
11:36:09.079 [PassFF.content] Inject icon login_password util.js:66:9
11:36:09.100 [PassFF.background] Message Page.goToAutoFillPending received in background util.js:66:9
11:36:09.238
Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist. background.js:3
```

[...]

Issue #387, opened 2019-02-12
Password loading fails when file contains a otpauth:// field


jarhill0:
### General information. Describe your environment

- Versions
  - Operating system: macOS 10.14.2 (18C54)
  - Browser: Firefox 65.0.1 (64-bit)
  - PassFF: 1.6.2
  - Host app: "testing" (not sure; pre 1.1.0)

Status line output: `[14:12:50] show -> (0) no error message`

- Output of `pass`: 
  ```
  [password censored]
  login: jarhill0
  url: https://github.com/login
  otpauth://totp/GitHub:jarhill0?secret=[secret censored]&issuer=GitHub
  ```
- Your preferences
  - In passff:
    - "OTP Auth field names": otpauth
  - In the host app `passff.py`: 
  ```
  # Default command for MacOS:
  COMMAND = "/usr/local/bin/pass"
  #COMMAND      = "/usr/bin/pass"
  COMMAND_ARGS = []
  COMMAND_ENV  = {
      "TREE_CHARSET": "ISO-8859-1",
      # Default PATH for MacOS:
      "PATH": "/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin",
  }
  CHARSET      = "UTF-8"
  ```
- Messages in the web console: 

```
[PassFF.content] 3 matches for https://github.com/login/ util.js:66:9
[PassFF.content] Best fit item personal/github for https://github.com/login/ util.js:66:9
[PassFF.content] Inject icon login_field util.js:66:9
[PassFF.content] Inject icon password util.js:66:9
[PassFF.content] Runtime port has crashed: Pass.executePass Error: "Invalid Version: testing" util.js:66:9

    logPrototype moz-extension://34ea875d-5229-3d4f-a750-ba9916e699b3/modules/util.js:66
    background_exec moz-extension://34ea875d-5229-3d4f-a750-ba9916e699b3/modules/util.js:137 

```
---

### Actual behaviour
When I attempt to fill (or fill and submit, or display) the password, nothing happens and the error seen above is displayed in the console. This *only* happens for sites which have a line like `otpauth://totp/GitHub:jarhill0?secret=[secret censored]&issuer=GitHub` (OTP, not necessarily GitHub OTP), and it happens for all of them.

### Expected behaviour
The password should be filled (or filled and submitted, or displayed).

[...]

PR #393, opened 2019-02-24
Assume 'testing' is valid for otp, see #387 #389

5bentz:
@madduck 
@LucidOne


madduck:
I can confirm that it now works, even after resetting the version to testing.

madduck:
Thank you!!

PR #399, opened 2019-03-30
Prefs: Get shortcut from browser.commands, fix #397

### LINKED ISSUE #397: Changing keyboard shortcut through "Manage Extension Shortcuts" does not keep setting
### General information. Describe your environment

- Versions
  - Operating system: Debian buster/sid
  - Browser: Firefox 66.0
  - PassFF: 1.6.3
  - Host app: 1.1.0

---

### Actual behaviour

In Firefox 66, there is a new way to manage extension shortcuts from the `about:addons` page. When changing passff shortcut from there, the shortcut is reverted back to the one defined in the settings. If will only work once, then revert.

### Expected behaviour

The shortcut defined in "Manage Extension Shortcuts" should be synchronize or replace the one in passff settings.

tuxor1337:
In my tests, this fixes #397. Please confirm!

### DIFF
--- a/src/modules/preferences.js
+++ b/src/modules/preferences.js
@@ -121,24 +121,22 @@ PassFF.Preferences = (function () {
 
   var prefObj = {
     init: function () {
-      return PassFF.Preferences.getBrowserCommand()
-        .then((command) => {
-            prefParams['tbMenuShortcut'] = command.shortcut;
-            return browser.storage.local.get(Object.keys(prefParams));
-        })
+      let changes = {};
+      return browser.storage.local.get(Object.keys(prefParams))
         .then((res) => {
-          let obj = {};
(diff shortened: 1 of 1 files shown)


vonHabsi:
Hi, this question is not so much about the issue but I am a newbie to WebExtensions and I want to understand how the process of updating a shortcut changed in the "Manage Extensions Shortcut" in Firefox works.

Which database are they stored, or what API function reads them? When the values are read is the addon restricted to those which have been assigned to it or can it read the shortcuts for all the other addons?

Thanks

tuxor1337:
Hi, please see here: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/commands As far as I understand the docs, an extension can only access and manipulate its own dedicated shortcuts.

Issue #374, opened 2018-11-17
Use URL field in password data for matching


mbrunnen:
The URL field in the password data should be, when present, the only value for URL matching quality. It seems right now, only the file name is taken in consideration. That leads, at least in my case, to a lot of false positives.

I think [`getItemQuality`](https://github.com/passff/passff/blob/master/src/modules/pass.js#L152) is the according function.

tuxor1337:
Impossible: We cannot match on the url field or any other entry content, because this would require decrypting the whole password store (see also https://github.com/passff/passff/issues/360 and several other previous requests).

mbrunnen:
oki thanks.

sunaurus:
> We cannot match on the url field or any other entry content, because this would require decrypting the whole password store

@tuxor1337 can you go into a bit more detail about what exactly is the problem here? This is the behavior of `pass grep` anyway, which is built into pass itself, and if the behavior is not a problem for pass, why is it a problem for passff? 

If it's a question of performance, I'm sure there are plenty of users for who this would not be an issue. I tested for 2 different real pass databases, both have over 100 passwords: a FULL `pass grep <url>` takes ~5 seconds on both, and stopping after first match is of course on average much faster. 

To me an ideal solution would be something like running a full `pass grep` with `window.location.host` and sending all the results to passff as they come in (along with a loading spinner while grep is still running to indicate that the results are partial). This behavior could also of course be off by default so that only users with smaller password stores or faster computers can choose to enable it in the settings.

tuxor1337:
You are right. When pass supports something like this, we should at least have an open discussion about supporting it in PassFF, as well (at least as an optional feature). Reopening...

phryneas:
Just as a comparison, but my store is on the extreme side - ~650 entries and a smartcard for decryption:

```
% time pass grep test
[...]
pass grep test  3.08s user 1.45s system 1% cpu 4:18.48 total
```

So, if you add this, please don't make it a default ;)

mbrunnen:
pass should use git grep:
```zsh
% time git --no-pager grep --no-index test
git --no-pager grep --no-index test 0.56s user 0.32s system 313% cpu 0.278 total
% time grep -r --exclude-dir=.git test
grep -r --exclude-dir=.git test 0.55s user 0.12s system 99% cpu 0.673 total
```

Spheerys:
I'm also very interrested by finding a solution for greping the url string instead the filename for the url matching.
[...]

PR #402, opened 2019-04-13
Index all password store meta urls on passff init

sunaurus:
There are many situations in which a single pass entry corresponds to multiple completely different web urls. One way of handling this is to list all the urls a password corresponds to inside the password entry using an `url: ` meta tag. However, this is currently not supported by passff (#374). 

In the linked issue, the possibility of using `pass grep` to find these meta urls was discussed. This would generally require decrypting the entire password store on every login, which might be acceptable for some password stores with fast hardware, but probably not a very good experience when using for example a smartcard to decrypt.

What I'm instead proposing in this PR is to only decrypt the entire store once on the initial load of the addon and to create a `Map` of all meta urls contained in the store. This index of meta urls would live in memory and would have a structure like:

```
{
  'services/amazon': ['amazon.com', 'amazon.co.uk', 'audible.com'],
  ...
  'fitness/impuls.ee': ['impuls.ee', 'arcticsport.ee']
}
```  

If this index exists, then the `hostMatchQuality` function can use it as an additional way of checking for matching entries. For example, my amazon entry is located at `services/amazon`, but it's still working on audible.com:

Screenshot of passff

The trade-off with this approach is that the list of pass entries and corresponding urls is now sitting decrypted in the user's browser memory, however, I believe this is an acceptable trade-off for a lot of users, and as long as this functionality is kept optional and explained next to the setting to turn it on, I think this is the best solution to #374.

This PR is not complete, I would still need to add the corresponding setting to toggle this functionality. I just wanted to submit it early to get some feedback. Does this approach seem reasonable?

This depends on passff/passff-host#38

### DIFF
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -128,6 +128,9 @@
   "passff_prefs_url_field_names_label": {
     "message": "URL field names"
   },
+  "passff_prefs_index_meta_urls_label": {
+    "message": "Index URL fields on startup. This will read all URL field values into browser memory and allow them to be used when passff looks for passwords. (Requires passff-host >=1.2.0)"
+  },
   "passff_prefs_otpauth_field_names_label": {
     "message": "OTP Auth field names"
   },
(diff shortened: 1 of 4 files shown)


tuxor1337:
I didn't look at the code, but the idea sounds okay to me as long as it's an optional feature that's disabled by default.

sunaurus:
Great! I added a setting with a description of how it works into the preferences page (disabled by default). Sadly, I don't know German, French or Russian, so I can't translate it for the other messages.json files. Other than that, this PR should be ready for code review.

tuxor1337:
Nice, but there is still one thing to worry about: The Firefox extension will be upgraded automatically. But the host app won't. Do you force users to upgrade their host app manually even if they don't need the new feature? Otherwise you have to deactivate this feature (and make note of it in the preferences) if the user doesn't have the required host app version...

sunaurus:
Will the next version of passff-host be 1.2.0? I added a note to the preferences page that the indexing option requires a passff-host >=1.2.0 and I also added a warning when running the `grepMetaUrls` command with a version lower than 1.2.0 (as was done with the `otp` command).

PR #410, opened 2019-05-21
Implement blacklisting of specific files/folders

### LINKED ISSUE #401: Regex white-/blacklist of paths to include
I have things in my password store which are not passwords, and I would like it if the fuzzy search didn't accidentally include them. I generally keep all of my web based passwords under the folder `web/`, so for me, this feature would ideally be a series of regex's to match against when you refresh the password list which are checked against for inclusion, such as:

- `^web/`
- `^otp/`

Or by default it can catch everything with:

- `.`

Or be toggleable. I will try to implement this feature myself, but familiarizing myself with the code will take time, and if it seems like this would be an easy feature to implement, then it doesn't hurt for me to ask.

cole-h:
The current implementation is just a blacklist, but I imagine it could
easily be extended to being a whitelist based on a toggle. If that is
desirable, the easiest way to do this would be to create a radio
button (checkbox might work as well, but would be less clear) and check
its state in `isHidden()`. If it is in blacklist mode, `return match`.
Else, `return !match`. Easy inversion.

I'm completely unfamiliar with JS, so if anything could be achieved in a
more concise manner, please feel free to tell me and I'll fix it right up.

---

Might fix #401

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ Some of them are described below:
   - Password field names
   - URL field names
   - OTP Auth field names
+  - Regex for hiding items (tests the full path to the item)
 - Adding Passwords
   - The default length for generating passwords
   - Whether or not to include special characters in generated passwords by default
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -389,6 +389,7 @@ PassFF.Pass = (function () {
               isMeta: null,
               hasMeta: null,
               fullKey: (curDepth === 0) ? key : curParent.fullKey + '/' + key,
+              isHidden: null,
               children: []
             };
 
@@ -403,6 +404,8 @@ PassFF.Pass = (function () {
               rootItems.push(item);
             }
           });
+
+          var isHiddenRegex = new RegExp(PassFF.Preferences.filterPathRegex.join("|"), 'i');
 
           allItems.slice().reverse().forEach(item => {
             let siblings = rootItems;
@@ -419,6 +422,7 @@ PassFF.Pass = (function () {
                                            || isUrlField(item.key)
                                            || isOtpauthField(item.key));
             item.hasFields = item.children.some(c => allItems[c].isField);
+            item.isHidden = isHiddenRegex.test(item.fullKey);
           });
 
           return [allItems, rootItems];
@@ -630,7 +634,8 @@ PassFF.Pass = (function () {
         .map(i => getItemQuality(i, urlStr))
         .filter(i => (i.quality >= 0))
         .sort((i1,i2) => (i2.quality - i1.quality))
-        .map(i => i.item);
+        .map(i => i.item)
+        .filter(i => !i.isHidden);
       log.debug(matchingItems.length, 'matches for', urlStr);
       return matchingItems;
     },
--- a/src/modules/preferences.js
+++ b/src/modules/preferences.js
@@ -89,6 +89,7 @@ PassFF.Preferences = (function () {
     showStatus            : true,
     tbMenuShortcut        : '',
     recognisedSuffixes    : 'co.uk,org.uk,me.uk,co.jp,com.au',
+    filterPathRegex       : 'notes\nattributes\nattachments',
   };
 
   var listParams = {
@@ -100,7 +101,8 @@ PassFF.Preferences = (function () {
     'urlFieldNames'       : ',',
     'otpauthFieldNames'   : ',',
     'autoFillBlacklist'   : ',',
-    'recognisedSuffixes'  : ','
+    'recognisedSuffixes'  : ',',
+    'filterPathRegex'     : '\n',
   };
 
   var lowerCaseParams = [
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -134,6 +134,9 @@
   "passff_prefs_otpauth_field_names_label": {
     "message": "OTP Auth field names"
   },
+  "passff_prefs_filter_path_regex_label": {
+    "message": "Regex for hiding items, one per line (tests full path)"
+  },
   "passff_prefs_script_howto": {
     "message": "See passff.py for changing the environment variables of pass."
   },
--- a/src/content/preferences.html
+++ b/src/content/preferences.html
@@ -117,6 +117,9 @@
           </p><p>
             <label>passff_prefs_otpauth_field_names_label</label>
             <input type="text" id="pref_otpauthFieldNames" value="" />
+          </p><p>
+            <label>passff_prefs_filter_path_regex_label</label>
+            <textarea id="pref_filterPathRegex"></textarea>
           </p>
         </fieldset>
       </div>


src/modules/preferences.js
```diff
@@ -89,6 +89,7 @@ PassFF.Preferences = (function () {
     showStatus            : true,
     tbMenuShortcut        : '',
     recognisedSuffixes    : 'co.uk,org.uk,me.uk,co.jp,com.au',
+    filterFieldRegex      : 'notes,att((ributes)|(achments))',
```
5bentz: I don't know why, the regex `notes|att((ributes)|(achments))` with 'gi' (global, case insensitive) doesn't match consistently the words "attachments" and "notes"

```javascript
var reg = new RegExp('notes|att((ributes)|(achments))', 'gi')
let truthy = [ "notes", "attributes", "attachments" ];
let falsy = [ "not", "at", "web", "pass", "notations", "att" ];

falsy.filter(e => reg.test(e)); 
// expected output: [ ]
// actual output: [ ]

truthy.filter(e => !reg.test(e));
// expected output: [ ]
// actual output: [ "notes" ] or [ "notes", "attachments" ]
```
Unsetting the global flag `g` (which seems useless anyway?) fixes the problem.

src/modules/pass.js
```diff
@@ -630,7 +641,8 @@ PassFF.Pass = (function () {
         .map(i => getItemQuality(i, urlStr))
         .filter(i => (i.quality >= 0))
         .sort((i1,i2) => (i2.quality - i1.quality))
-        .map(i => i.item);
+        .map(i => i.item)
+        .filter(i => !i.isFiltered);
```
5bentz: I think the state `isFiltered` quite misleading: we have to think about it: "Okay, it is filtered, so it is displayed, isn't it? Oops"
I would prefer `isVisible`.
* isFiltered -> isVisible
* isInFilter -> isVisible

On the other hand, I would keep the parameter "filterFieldRegex" because this could be a blacklist or a whitelist.

5bentz:
Hey cole-h,
thanks for the contribution!

I'm not a fluent JS either, but I tried to help a bit :+1:

README.md
```diff
@@ -145,6 +145,7 @@ Some of them are described below:
   - Password field names
   - URL field names
   - OTP Auth field names
+  - Filter fields out by regex (blacklist)
```
5bentz: Huh, I'm quite tired, but I'm pretty sure we are dealing with objects called filenames, password names, passwords or items.
Fields are found in the password data ("login" "url "http" are commonly used fields)

src/modules/pass.js
5bentz: I just realized the entries can also be filtered by the context : see the "Display contextual" button in the toolbar menu.

`isHidden` could be an alternative.

src/modules/preferences.js
cole-h: Good catch, thanks. Not familiar enough with regexes to tell you why that happens, either, but I'll remove the global flag. If you can provide a convincing enough reason to remove the case-insensitive flag, I'll do so as well -- however, I'd rather `attachments` and `Attachments` both match, in case a user capitalizes a letter by accident.

src/modules/pass.js
cole-h: I like it. Will show up in my next push.

README.md
cole-h: You weren't tired, but I was. Thanks for all the feedback. I'll probably go over my changes once more tomorrow.

src/modules/preferences.js
norcalli: From the description of `test`:

```
If the regex has the global flag set, test() will advance the lastIndex of the regex. A subsequent use of test() will start the search at the substring of str specified by lastIndex (exec() will also advance the lastIndex property). It is worth noting that the lastIndex will not reset when testing a different string.
```

The method `test` mutates the Regex, and since you aren't reinstantiating it for each iteration of the loop, you are getting mixed results. Your test is, therefore, the problem, and not the regex.

Regardless, the `g` flag is inconsequential for the `isHidden` method, since we only case about finding a single match. Therefore, it is unnecessary.

Additionally, a proper regex engine will do the work of compiling `notes|attributes|attachments` into a DFA which will be equivalent to writing `notes|att(ributes|achments)` because it will collapse the common prefixes. Therefore, for clarity, I would recommend making the default regex `notes|attributes|attachments` for improved clarity.

src/modules/preferences.js
```diff
@@ -100,7 +101,8 @@ PassFF.Preferences = (function () {
     'urlFieldNames'       : ',',
     'otpauthFieldNames'   : ',',
     'autoFillBlacklist'   : ',',
-    'recognisedSuffixes'  : ','
+    'recognisedSuffixes'  : ',',
+    'filterFieldRegex'    : ',',
```
norcalli: This could potentially be error prone. Since `|` acts as a natural separator in regex, essentially all this does is replace `,` with `|` in the final result. And, as unlikely as it may seem, someone may use a `,` in their regex, which will cause a bug. Is it possible to make this field *not* a list, and just input a raw regex?

src/modules/pass.js
```diff
@@ -419,6 +429,7 @@ PassFF.Pass = (function () {
                                            || isUrlField(item.key)
                                            || isOtpauthField(item.key));
             item.hasFields = item.children.some(c => allItems[c].isField);
+            item.isHidden = isHidden(item.fullKey);
```
norcalli: While the method is nice for separating concerns, for performance reasons, you could avoid recreating the regex for every single item by creating it once right before this block starting at `allItems`.

Something to the effect of:

```js
var isHiddenRegex = new RegExp(PassFF.Preferences.filterFieldRegex, 'i');
allItems.slice().reverse().forEach(item => {
  // ...
  item.isHidden = isHiddenRegex.test(item.fullKey);
  // ...
});

src/modules/preferences.js
cole-h: You're right; I didn't take that into consideration. I'll fix this in my next push, addressing this and your next comment.

Then, I'd like to add a note about how the regex needs to be separated by pipes for each different match. Do you have any ideas on where this should go?

Under my addition to the README as a note?
Under the actual entry in the preferences? And if you think that is preferable, what would this look like? I'm completely unfamiliar with writing HTML.

src/modules/pass.js
cole-h: Smart. Something I should have noticed myself. Cheers.

src/modules/preferences.js
norcalli: This might be able to be done with just the default field value and the label.

The default field value will be something like `notes|attributes|attachments` which should answer the question of "How do I target multiple items."

And the label can make it clear that it is a single regex which tests against the full path of the entry. Something like `Single regex for hiding items (tests against the full path):`

cole-h: A fairly simple solution. Guess I'll just have to hang out in the issues to make sure it's clear to users. Thanks for your input! Push should show up soon. Hopefully everything has been addressed.

tuxor1337: Having only a single regular expression for all kinds of tests is a bit tedious in case one of the tests is supposed to test from the beginning of the path, i.e., using `^` (or, same problem: `$`). Maybe we could instead have a multiline textfield and have one test per line?

cole-h: @tuxor1337 Before I push, is this what you had in mind? https://gist.github.com/cole-h/1cdacc1d89a722869d405e80c41ade21 (diff between current branch HEAD)

tuxor1337: Yes, that looks okay. Just one remark: the `<textarea>` isn't a *void element* according to the HTML5 specification (https://w3c.github.io/html/syntax.html#void-elements). So it can't be self-closed using a slash `/` in the start tag but it has to be followed by the corresponding end tag `</textarea>`. Please change this before pushing.

cole-h: Whoops. Thanks for catching that. As you can tell, I don't use HTML much.

norcalli: I have nothing to add other than saying the textarea is a good compromise, the implementation looks good, and good on you for making this PR out of your comfort zone.

norcalli:
Can we merge this? I've been using it for the past 2 weeks, but I have to reinstall it whenever I restart firefox. It's working very well.

tuxor1337:
One last suggestion. I find the variable name `filterFieldRegex` as well as the message name `passff_prefs_filter_field_regex_label` extremely misleading since this is about filtering entries according to their path and not about filtering any fields (according to the nomenclature that we're using here). So, would you please change this to `filterPathRegex` and `passff_prefs_filter_path_regex_label` or similar?

cole-h:
Oops, I missed those two; thanks. New (and hopefully final :stuck_out_tongue: ) commit should show up soon.

norcalli:
Just gonna ping this again.

tuxor1337:
Thanks to you all and a special thanks for @cole-h for your contributions :)

cole-h:
Thanks for being patient with me 😊 Additional thanks for releasing a new version with this added!

PR #434, opened 2020-03-11
Add menu width preference

runejuhl:
This PR adds customizable width of the menu popup The default menu width set to `28rem` is now customizable through the addon preference page:

passff-width


tuxor1337:
Looks good, thanks for your contribution! However, can you explain why you changed the default width? I would prefer to leave the UI unchanged unless there are good reasons for a change.

runejuhl:
> However, can you explain why you changed the _default_ width? I would prefer to leave the UI unchanged unless there are good reasons for a change.

Sure -- it's done for purely selfish reasons. I have multiple git trees that are symlinked into my `~/.password-store` folder, making for some really long names, and the default (which spurred all of these changes) is simply to narrow.

I figure that users today have higher display resolutions compared to when passff was first created, and that I'm not the only one seeing this issue.

I'd be fine with reverting that change though, just let me know :slightly_smiling_face:

tuxor1337:
Hm, I understand your reasons, but I'm not convinced, judging from my own subjective personal experience. :-P If you don't mind, please reset the default value to 28rem, and I will be happy to merge this.

runejuhl:
>  If you don't mind, please reset the default value to 28rem, and I will be happy to merge this.
Sure thing -- done! :heavy_check_mark:

tuxor1337:
Okay, great! Give me a few days. I will probably make a new version out of this and #435.

runejuhl:
Nice! Let me know if you want me to change anything. ...and thank you for passff! :slightly_smiling_face:

runejuhl:
Just noticed that I've missed some CSS changes for a dialog; I'll try to get it fixed soon.

buggy-css

src/skin/menu.css
```diff
@@ -180,10 +234,26 @@ body.error > div {
...
+  color: #000;
+}
+
+#statusbar, .itemPickerTarget, .results select {
+  border-color: #b6b6b3;
+  border-style: solid;
+}
+
+button, input, #statusbar, .itemPickerTarget, .results select {
+  background-color: inherit;
```
tuxor1337: Why do you add this? Maybe you only want to set this for the `body.dark` case?

tuxor1337:
Looks fine now, thanks for applying the proposed changes!

PR #435, opened 2020-03-11
Lift `rateLoginInput` heuristics from `ratePasswordInput`

### LINKED ISSUE #433: Email field with blank name is not detected
### General information. Describe your environment

- Versions
  - Operating system: Linux
  - Browser: Firefox Nightly 74.0a1
  - PassFF: 1.9.1
  - Host app: 1.2.1

Status line output: `[18:27:48] show -> (0) no error message`

This is likely due to the fact that the Discord email field has an empty `name`. Manually changing it to be `name="email"` allows it to autofill fine. However, the `type` is not blank, and set to `"email"`, so maybe we could use the input's `type` as an additional heuristic for autofill?

---

### Actual behaviour
When attempting to autofill a Discord login, the email field is left blank.

### Expected behaviour
The email field should have a PassFF button.

### Steps to reproduce the behaviour
1. Navigate to https://discordapp.com/login
2. Notice the lack of PassFF icon in the email field
3. Type it in manually

cole-h:
My use case is: for some reason, the email field is not detected on
Discord. This means I have to either 1) type in my email or 2) inspect
element and change the empty `name=""` to something from my login input
list. This change directly copies the implementation of
`ratePasswordInput`. I tested on Discord, GitHub, and Google login pages
and everything still appears to work.

---

I didn't inspect my changes thoroughly and only copy-pasted from directly above in `ratePasswordInput`. I don't know if returning 100 is what we want, but it seems to do the trick. Let me know if there's anything I failed to consider.

---

Fix #433

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -128,11 +128,12 @@ PassFF.Page = (function () {
   }
 
   function rateLoginInput(input) {
-    if (loginInputTypes.indexOf(input.type) < 0) {
-      return 0;
-    } else {
+    if (input.type === 'email') {
+      return 100;
+    } else if (input.type === 'text') {
       return rateInputNames(input, PassFF.Preferences.loginInputNames);
     }
+    return 0;
   }
 
   function rateOtpInput(input) {


tuxor1337:
Okay, thanks for the PR!

I won't be able to just accept this since the fact that the variable `loginInputNames` is ignored makes me very suspicious. For example, this will break login forms with fields of type `tel`. Furthermore, it feels like this is not the most elegant solution for your problem. I hope I find time to think this over in the next couple of days.

cole-h:
I figured it wasn't the best way to achieve this, but I might as well leave the result here for others (or you) to find a more elegant solution.

I'm running an unsigned version with this patch for now, so even if you don't accept it, no harm, no foul. No need to rush; it works for now, though a better solution would be nice when you have the time ;)

tuxor1337:
I now decided to fix it with a different approach. Thanks anyway for the push :+1:

cole-h:
Confirmed it works, thanks!

Issue #441, opened 2020-06-04
Suggestion: alternate default action on Enter key


mildred:
### Use case

In some sites, there is no standard login/password, but there is a login, a secret key and a password. On others (banks mostly), the auto filling does not work because of convoluted login process. In those cases, the easiest way to login is to display the password-store entry. This is fine except there is no easy way to display the entry using the keyboard.

Example scenario:

- `<C-y>example.com`
- up/down to select entry
- `<Right>` (and not `<Enter>` which would trigger auto filling which does not work)
- `<Down><Down><Down><Down><Down><Down><Down>` to select Display
- `<Enter>` (and not `<Right>` which moves to the top of the menu)

Also, the display action in the sub menu is not displayed by default because the menu is too short.

### Proposal

1. Make it so that the `<Right>` key has the same action as the `<Enter>` key when in the actions menu. It's counter-intuitive to have different keys to activate a menu entry depending on the menu level

1. Resize the menu a bit so that the display action is visible and no scrolling is needed.

1. In the preferences, allow the *Behavior of Enter key* to map to the display action

1. Allow alternate activation key mapping such as `<Ctrl-Enter>` or `<Shift-Enter>` and allow them to configure to the display action if desired. By default, one could map to Fill only, and the other to Display

tuxor1337:
Oh, short comment on your proposal 1: This is #436 and a fix will be included in the upcoming release.

PR #442, opened 2020-06-04
Allow to configure alternate Enter key behavior

mildred:
Implements #441

Implements various new behaviors on Enter key, and allow to configure Shift-Enter and Ctrl-Enter too. Shift-Enter was set to open in a new tab when the behavior was *go to* and it is still the default behavior but it is now configurable.


tuxor1337:
Really nice feature, clean implementation, including locales for several languages. Thanks for your contribution!

PR #469, opened 2021-02-05
Skip inputs with placeholder containing `search`

eyalzek:
such as Gmail's search bar.

Another issue that I noticed (which is solved by this PR in gmail's case, but might be present elsewhere), is that with the current version of passff, the pass icon on the gmail search bar is repeated (`background` attribute is set to `repeat`):
image


tuxor1337:
Thanks for the easy fix!

PR #486, opened 2021-07-22
Add supported fields in README

xeruf:
By the by, why do you only use 3rd and 5th-level headlines?
Seems as if you are trying to hack the output to your preferences within the github renderer, which is exactly what markdown was supposed to avoid ^^


tuxor1337:
This is obsolete due to 4a8a8d4

xeruf:
and all the other improved tidbits you ignore? :c

Issue #343, opened 2018-06-22
Multiple HTTP auth dialogs are spawned at the same time


14mRh4X0r:
### General information. Describe your environment

- Versions
  - Operating system: Arch Linux
  - Browser: Firefox 60.0.2 (64-bit)
  - PassFF: 1.3
  - Host app: 1.0.1

Status line output: `[10:52:12] ls -> (0)`

- Your preferences
  - In PassFF: not changed from the default
  - In the host app `passff.py`: not changed from the default
- Messages in the web console: none

---

### Actual behaviour
When opening a cached page that uses HTTP authentication, each parallel request to the server (images, javascript, etc.) spawns a new PassFF dialog, potentially spawning hundreds.

### Expected behaviour
PassFF should display at most one HTTP auth dialog at a time.

### Steps to reproduce the behaviour
1. Go to a cached web page with multiple images/javascript files that uses HTTP authentication
2. Observe that for each request, a PassFF dialog is spawned

For me, this is especially noticeable with a [Netdisco](http://netdisco.org/) install.

This issue might be related to #322.

tuxor1337:
Can we somehow check whether two auth requests are exptected to be resolved by the same credentials?

shtrom:
Sounds like #322

arthaud:
I also have this problem (spawns 11 dialogs for a specific url). I would love to see a fix.

shtrom:
Still happening here. It actually killed my Firefox with 500+ windows yesterday :laughing: 

As I said before, this seems to happen on cached pages, where FF tries to re-fetch all the objects at once, and gets separate 403 for each.

I'm wondering whether the number of concurrent popups could be limited by, say, one per combination of hostname and HTTP Auth Realm. This should do the trick to limit the requests for resources on the same server.

shtrom:
Ok, I got a small hacking reproduction case for this. Essentially, it's just a small express-based server that asks for HTTP Auth on anything but the index. The index has two images in it.
[passff-auth.zip](https://github.com/passff/passff/files/6967978/passff-auth.zip)

When hitting http://localhost:3000, passff spawns one pass-entry window per image.
![Screenshot from 2021-08-11 21-28-07](https://user-images.githubusercontent.com/160280/129021345-38211db4-0776-4f5d-8dee-82f3a53c7a5e.png)

tuxor1337:
[...]

Issue #322, opened 2018-04-04
Handle multiple HTTP auths in parallel


shtrom:
### General information. Describe your environment

I have three pinned app tabs to subdirectories of the same server using the same HTTP Auth. Firefox asks for HTTP Auth separately for all three. When HTTP Auth support is enabled and Firefox starts, PassFF keeps popping up new, empty, dialog boxes (`moz-extension://...`) that overlap each other and the HTTP Auth windows, making it hard to enter the credentials. As the Passff's dialog boxes are empty, it is also useless in this situation.

Note that the empty dialog boxes seem to be similar to the empty dialog boxes I get when I try to `Display` the content of an entry---most of the time.

I guess it's a race condition where Firefox tries to actively pre-load all three tabs, and has to display 3 parallell HTTP Auth dialog, which means that PassFF has to do the same thing.

- Versions
  - Operating system:  ArchLinux x86_64
  - Browser: Firefox 59.0.2
  - PassFF: 1.3b (and all previous since HTTP Auth support was added)
  - Host app: 1.0.1

Status line output: `[20:03:16] show -> (0)`

- Your preferences
  - In PassFF: Handle HTTP Auth, Log to JS console
- Messages in the web console: nothing useful \:

---

### Actual behaviour
Firefox displays HTTP Auth windows one after the other; PassFF keeps popping up more `moz-extension://` windows, with no content, that overlap each other and the HTTP Auth dialog.

### Expected behaviour
Firefox displays the first HTTP Auth window, which gets handled by PassFF with its (hopefully non-empty) popup, then move on to the next, handled the same way, untill all pinned app tabs are logged in or cancelled.

### Steps to reproduce the behaviour
1. Pin 3 sites requiring HTTP Auth as app tabs
2. Restart Firefox

tuxor1337:
The "empty" popup windows should have been fixed already: https://github.com/passff/passff/issues/257 But for some reason, the fix stopped working. I will commit a new workaround for that in a moment.

shtrom:
Wow. Cheers for the quick fix!

shtrom:
Hum, I just tried with the latest version from git. Installed in `addons:debugging` from the manifest.json, with the matching host script, but I still see a similar issue:
* it only opens 3 windows at a time, but seems to then be closing them then reopening them, at which point they sometime steal the focus,
* the window displayed is still empty.

tuxor1337:
[...]

PR #489, opened 2021-08-11
HTTP Auth: limit the number of popups

shtrom:
This change prevents from opening a large number of popups for embedded
objects requiring HTTP Auth (e.g., public or cached page data).

We do so by searching the currentAuths array for existing entries that
share the same Root URI and the same realm. If one is found, we don't
open a new popup, but return the Promise associated with the first auth
instead.

fixes: #343, #322

Signed-off-by: Olivier Mehani <shtrom@ssji.net>


tuxor1337:
Thanks for the PR! I merged and did some clean up in b2ae285

PR #490, opened 2021-09-08
Configurable Submit Button Queries

cs8898:
# What

- Added TextArea to preferences to edit button query strings (Seperated by \n)
- Dropped the Static List of Querys and fetch them from Preferences

# Why

- Needed for an Vaadin Application using Custom HTML Tags and JSF AJAX (So a fallback Form submit is not working)

# See

https://github.com/F43nd1r/Acrarium/issues/290


tuxor1337:
Thanks, looks great!

Would you be willing to share your custom query string? Maybe it's also beneficial for other users.

cs8898:
The Software i used is [Acrarium](https://github.com/F43nd1r/Acrarium), its a self hosted version of the Google Play Crash-Reporting Software.

It uses the "Standard" Vaadin LoginForm
https://vaadin.com/components/vaadin-login...

The query i use is `vaadin-button[part=vaadin-login-submit]`, thats the element holding the `onclick` listener.

So in total my current list is:

```txt
button:not([type=reset])
input[type=submit]
input[type=button]
vaadin-button[part=vaadin-login-submit]
```

tuxor1337:
Thanks! Does this work for you?
```
button:not([type=reset])
input[type=submit]
input[type=button]
[role=button]
```
If it works, maybe you can add this query string to the default query strings.

cs8898:
Yes the list works also for me...

I think the rule isn't the best because it lacks the node type "button" or "input", but it works on several other pages (Facebook, Gitea, Jenkins, Amazon and KeyCloak)

For Vaadin the node type has to be "vaadin-button" because there is a shadow-dom...

## TL;DR
I will add your list to my PR ;D

tuxor1337:
Thanks!

PR #503, opened 2021-12-15
adding otp name defaults for microsoft and gitlab

casonadams:
login.microsoftonline.com: `otc`
gitlab.com: `user[otp_attempt]`

### DIFF
--- a/src/modules/preferences.js
+++ b/src/modules/preferences.js
@@ -64,7 +64,7 @@ PassFF.Preferences = (function () {
   var prefParams = {
     passwordInputNames    : 'passwd,password,pass',
     loginInputNames       : 'login,user,mail,email,tel,username,opt_login,log,usr_name',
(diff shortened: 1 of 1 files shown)


tuxor1337:
Actually, I don't want to come in the situation, where I have to add every single input name of every OTP provider out there to the list. However, it seems like gitlab and microsoft are among the major OTP services and there won't be too many of comparable relevance, right?

casonadams:
Should I close this MR out?

tuxor1337:
No, it's cool. We will merge this for now, considering the relevance of those particular OTP providers. Still, your idea about a personal config file to keep under source control sounds much more elegant. For example, we could have a particular entry in your password store that is recognized as a PassFF config file. If you are still interested in such a feature feel free to start a conversation in a new GitHub issue, or directly propose your own implementation in a PR.

PR #512, opened 2022-07-28
Bugfix: you can now leave regex for hiding items empty.

nottoday888:
There is a bug where the top level view is enforced instead of the contextual view. This will occur when the field “Regex for hiding items” is empty (or only empty lines). Empty regex's will always pass Regex.test(). So, all items will be hidden, which is obviously not intended behavior.

I think I fixed it, but embarrassingly enough, I have trouble building and installing the extension (and so my patch) to test it. If someone could give some instructions, I would like to test it any ways.

### DIFF
--- a/src/modules/pass.js
+++ b/src/modules/pass.js
@@ -438,6 +438,7 @@ PassFF.Pass = (function () {
             }
           });
 
+          var isInUseHiddenRegex = PassFF.Preferences.filterPathRegex.length != 0;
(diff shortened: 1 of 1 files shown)


tuxor1337:
Here you find developer instructions: https://github.com/passff/passff/blob/master/docs/CONTRIBUTING.md#development

However, I tested your changes and confirm that it fixes the problem. Thanks a lot for your contribution!

PR #514, opened 2022-08-28
Support Cloudflare otp field name

rhubinak:
The field name for Cloudflare [two-factor authentication](https://dash.cloudflare.com/two-factor) is twofactor_token. Adding it to the list of otp fields.


tuxor1337:
As mentioned in #503, I absolutely want to avoid the situation, where I have to add every single input name of every OTP provider out there to the list. Users can manually add this kind of entry to the list already...

rhubinak:
Ah, apologies, haven't read that one. Makes sense, closing this.

PR #515, opened 2022-08-29
Fix infinite loop on text field input

rhubinak:
Repro steps (Firefox 104.0):
* Open a page with an input text field, like some issue from https://github.com/passff/passff/issues and focus (click inside) the comment input text field
* Open the debugger (CTRL+SHIFT+I) and see recurring `Empty string passed to getElementById()` error.

On some text fields, this leads to noticeable slowdown (Zendesk). Found out this is due to MutationObserver notifying on attributes in infinite loop in some pages. Setting this to false fixed this issue. Tested on ~100 different sites (~20 with otp).

Also noticed, that the debug console is complaining about a null being returned instead of an empty Promise, so fixed that as well.

### DIFF
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -529,7 +529,7 @@ PassFF.Page = (function () {
     bestFitItem = PassFF.Pass.findBestFitItem(matchItems, url);
 
     var obs = new MutationObserver(onNodeAdded);
-    obs.observe(document, { attributes: false, childList: true, subtree: true });
+    obs.observe(document, { attributes: true, childList: false, subtree: true, attributeFilter: ["id", "name"] });
     onNodeAdded();
 
     return PassFF.Page.goToAutoFillPending()
--- a/src/modules/page.js
+++ b/src/modules/page.js
@@ -78,6 +78,7 @@ PassFF.Page = (function () {
   }
 
   function readInputNames(input) {
+    log.debug("test")
     let inputNames = [input.name || "", input.id || ""];
     let placeholder = input.getAttribute('placeholder');
     if (placeholder && placeholder.toLowerCase().indexOf('search') === -1) {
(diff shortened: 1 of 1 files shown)


tuxor1337:
I'm also on Firefox 104.0 and don't get the message you mention in the Javascript console. Since attribute changes might be very relevant for what we try to achieve here, I would be against disabling it in the `MutationObserver`.

Furthermore, I don't even know which call to `getElementById` in PassFF might have caused that message. 

Are you using any other Firefox extensions in parallel?

rhubinak:
It shows that the error stems from function readInputNames in page.js, line 94

```
 92     // labels are <label> elements whose `for`-attribute points to this input
 93     if (input.labels) {
 94       inputNames = inputNames.concat(Array.from(input.labels, l => l.innerText));
 95     }
```

I will retest without extensions, it's possible that they are affecting the way that the DOM is generated.

tuxor1337:
> It shows that the error stems from function readInputNames in page.js, line 94

There is no call to `getElementById` in line 94?!

rhubinak:
I know, but that's what the debug console shows.

rhubinak:
Hmm, unable to reproduce on GitHub, but on Zendesk always, even with extensions off. Unfortunately, the issue is back even with my changes, so hunting some more.

rhubinak:
This is the second issue with the incorrect return value. Try reloading this page with the debug console on

rhubinak:
childList=false and attribute filter. This is the winning combination! I will test it some more in the coming days. I understand if this change proves too big or breaking for the upstream and I am willing to maintain my own fork. Thank you for all the hard work, Thomas!

tuxor1337:
With `childList=false` we won't be observing the addition of new elements. If new forms are created, we won't notice it, right?

However, it might make sense to add an attribute filter. Furthermore, it might boost performance to let the callback function `onNodeAdded` iterate through the mutations and apply some logic to determine whether any of the mutations are relevant for PassFF. I would expect that it can be quite challenging to define whether a mutation is relevant or not.

rhubinak:
Eh, workarounded by adding `zendesk` into autofill-ignore field. Should we pursue just the change in src/modules/page.js, line 793 to fix the debug console complaint? Probably makes sense to open a new PR for this and close this one, right?

tuxor1337:
Great that you were able to find a fix for the zendesk-issue! To shortcut this, I just pushed your fix for the `null` return value as [e1174c9](https://github.com/passff/passff/commit/e1174c97a5c85492d282f8b39fcdfb5d0e1bd489).

Issue #519, opened 2022-10-28
Option to prevent autofill if a subdomain matches only


st31ny:
### Versions
- Operating system: Debian Testing
- Browser: Firefox 106.0
- PassFF: 1.14
- Host app: 1.2.1

Status line output: `[07:31:31] show -> (0) no error message`

### Observations

- Your preferences
  - In PassFF: autofill enabled 

---

### Steps to reproduce the issue
1. Have an entry in the db with a certain subdomain test.example.com.
2. Surf to another website with the same subdomain test.foobar.org.

### Actual behaviour

Login credentials were autofilled.

### Expected behaviour

I would prefer to have an option to enable a stricter domain checking in autofill as proposed by @thblt in #199 before (apparently, his proposal was not implemented in the end). Since test.example.com and test.foobar.org just share a subdomain, there is no point in filling the credentials for test.example.com at test.foobar.org. The fuzzy search is of course great for manual searches, but a potential security risk during autofill.

In my opinion, such an option, let's say "Strict domain matching for autofill" should behave roughly as follows:
1. If an `url` field is present in the entry and has a path: the entire path must match (query can be ignored).
2. If an `url` field is present in the entry and has no path (i.e., it's just the domain): match this domain and any subdomain.
3. If no `url` field is present in the entry: match use the filename as domain as proceed as in 2.

This should actually yield a solid security improvement while still being able to profit from the autofill feature.

tuxor1337:
I think that this check has already been implemented in #333 and is available in the preferences in the autofill section: "Warn about potential phishing sites".

st31ny:
Sorry for my late reply, somehow missed your comment.

> I think that this check has already been implemented in #333 and is available in the preferences in the autofill section: "Warn about potential phishing sites".

The "Warn about potential phishing sites" is very intrusive, unfortunately, since it always pops up a modal dialogue. Furthermore, I am not sure — are filenames actually considered as Domains in the security warning?

tuxor1337:
Please provide more specific examples with step-by-step instructions on how to reproduce. Then describe the current behavior with the "Warn about potential phishing sites" option enabled and then explain your desired behavior.

st31ny:
Sure.
First, none of the entries in my pass db have a `url` field set but I rather rely on filenames being identical to domain names.

[...]

PR #520, opened 2022-11-13
Autofill domain check

st31ny:
This adds an option to check for exact domain matches when auto filling. When the additional security checks are enabled, a failure simply prevents auto fill rather than popping up a modal dialog.
 #519


tuxor1337:
Thanks a lot, this is a great contribution!

From a programming point of view, it's a bit odd to have these two functions without one of them calling the other:

* `isSubdomainInclusive` checks if the second hostname is a subdomain of the first.
* `domainSecurityCheck` checks if both hostnames are subdomains of the same domain.

I think the function `domainSecurityCheck` should extract the domain from the first hostname and then use `isSubdomainInclusive` to check whether the second hostname is a subdomain of the first's domain.

From a usability point of view, I think that the two check-mark options `passff_prefs_autofill_subdomain_check_label` and `passff_prefs_domain_check_label` are very specific and unnatural. It's not very transparent to a user what they do, how they differ and why you would go for one or the other option. We should rather have these options and several combinations thereof:

* A preference to decide which of the following three things to do when a website is not considered trustworthy:
    1. *ignore* (just auto-fill, don't even do the security checks),
    2. *warn* (the user may override the security checks),
    3. *enforce* (the auto-fill is silently omitted if the security checks fail).
* A preference to decide which security checks to apply: 
    1. Require that the current website's URL is a valid URL (in the sense of JavaScript's `URL` feature).
    2. Requre that the URL stored in the password store is a valid URL (in the sense of JavaScript's `URL` feature).
    3. Require that the current website's hostname and the hostname stored in the password store are both subdomains of the same domain.
    4. Require that the current website's hostname is equal to or a subdomain of the hostname stored in the password store.
    5. Require that the security of the protocol (https or http) is at least as good as the one stored in the password store.
* A preference to decide whether the entry's key name is used as a hostname in the security checks if no `url` field is present.

Ideally, it would be possible for each of the 5 security checks separately to specify which of ignore/warn/enforce is applied if the check fails...

src/modules/page.js
```diff
@@ -621,6 +633,20 @@ PassFF.Page = (function () {
...
+    const passDomainParts = passDomainStr.split('.').reverse();
+    if(passDomainParts.length > currDomainParts.length) {
+      return false;
+    }
+    for (let i = 0; i < passDomainParts.length; ++i) {
+      if(currDomainParts[i] !== passDomainParts[i]) {
+        return false;
+      }
+    }
+    return true;
```
tuxor1337, line 647:
```suggestion
    const currDomainParts = currDomainStr.split('.');
    const passDomainLength = passDomainStr.split('.').length;
    return currDomainParts.slice(-passDomainLength).join(".") == passDomain;
```

st31ny:
> From a usability point of view, I think that the two check-mark options `passff_prefs_autofill_subdomain_check_label` and `passff_prefs_domain_check_label` are very specific and unnatural. It's not very transparent to a user what they do, how they differ and why you would go for one or the other option.

Yeah, I 100 % agree on this. I actually also needed to look into the source to actually understand what exactly is checked.

In addition, I also thought that we should probably consider `PassFF.Preferences.recognisedSuffixes` and require that the entry's key name consists of more than a TLD in the security checks.

I hope that I'll find some time today or tomorrow to have a look at the implementation again.

st31ny:
Sorry, didn't find the time earlier to finish the implementation. So far, I edited the settings page and added the new options for a detailed adjustments for checks. What do you think?

I think that requiring a valid URL can be implicit. When visiting a real website, the URL must be valid anyways and an invalid URL in the pass db can simply cause the check to fail but is not a security issue otherwise. Therefore, I only added the other three levels. Since ii, iii, and iv are getting in stricter in this order, I grouped them in a combo box and added another combo box for the independent setting of the protocol check.

The actual logic needs to be implemented still…

xeruf:
Hey, any news on this great contribution? :)

st31ny:
> Hey, any news on this great contribution? :)

I was quite busy in the last weeks but hope that I'll find some time at the next weekend or so. But good to hear that others want to see this, too :).

tuxor1337:
@st31ny Since people occasionally complain about the warnings caused by the security checks (e.g. https://github.com/passff/passff/issues/564), this PR would still be very welcome. Do you still have plans to wrap up the work?

st31ny:
@tuxor1337 , sorry that I did not manage to come back to this in such a long time. Do you think the concept still makes sense?

I cannot promise that I'll find some hours in the next weeks, please feel free to build upon my PR.

tovogt:
Yes, I think the concept still makes sense. I just implemented the feature discussed in this PR as far as I understand it.

It would be great if you found the time to test this implementation and provide some feedback. :)

tuxor1337:
Yes, I think the concept still makes sense. I just implemented the feature discussed in this PR as far as I understand it.

It would be great if you found the time to test this implementation and provide some feedback. :)

tuxor1337:
Okay, I think I tested this myself sufficiently to approve and merge this.

st31ny:
Thanks a lot! It was still on my todo list, but hadn't found time yet!

st31ny:
Hmm, apparently, using an entry for a parent domain, does not work anymore.

So, the credentials for example.org are not used anymore on a site sub.example.org and cause a security warning if I do it manually.

tuxor1337:
So you have "warn & ask" for "If the site and the stored URL are different"? Then it's absolutely correct behavior if there is a warning when trying to fill the credentials for example.org while on sub.example.org.

st31ny:
> Your setting is "warn & ask" for "If the site and the stored URL are different"? Then it's absolutely correct behavior if there is a warning when trying to fill the credentials for example.org while on sub.example.org.

Oh, right, sorry. I think, I misread the setting.

So, the option "If the site and the stored URL are different" actually requires that the URL is an exact match, correct?

tuxor1337:
Yes I know, it can quickly be misread. Sorry for that, a clear language in preferences is not trivial - even more so if you are not formulating in your native language.

st31ny:
> Yes I know, it can quickly be misread. Sorry for that, a clear language in preferences is not trivial - even more so if you are not formulating in your native language.

No worries, I actually do not have an idea how to formulate this better, either…

st31ny:
But thanks a lot for finishing this — this has improved my experience with the addon *a lot*! (I.e., I could now re-activate autofill).

PR #526, opened 2023-02-15
Add Brazilian Portuguese translation

ftapajos:
Adds pt_br (or simply pt) locale


tuxor1337:
Thanks for the great contribution! :tada:

PR #548, opened 2023-08-09
Autofill lonely OTP inputs

VVishion:
OTPs are often challenged on a new page after the username and password was entered. This change allows this workflow.

This should not impose additional security risks given that OTPs are:
* only valid for a short amount of time
* only valuable in combination with an identity (e.g. login, session, ...)


tuxor1337:
Thank you for your contribution!

PR #568, opened 2025-08-14
feat(appearance): add offset to move the icon

wiebel:
Makes the icon in the input field movable. Handy if there are multiple password managers required.


wiebel:
PR in codeberg, sorry

